PR #195, opened 2016-03-31
Default account should only be selected onCreate, not onResume.

riaanjvos:
- After scanning private key for watch-only spend confirmation, selectDefaultAccount() was called after returning from qr scan activity, which updated the coin selection just before transaction signing needed to happen.


kristovatlas:
utACK

PR #285, opened 2016-08-10
Send refactor

riaanjvos:


app/src/main/java/info/blockchain/wallet/view/SendActivity.java
```diff
@@ -0,0 +1,744 @@
...
+                ;
+            }
+        };
+
+    }
+
+    @Override
+    public void onShowPaymentDetails(PaymentConfirmationDetails details, boolean isLargeTransaction) {
+
+        runOnUiThread(() -> {
```
ditn: I don't think needs to be in a specific UI thread - it's never called from background AFAIK. Same with a couple of other calls

app/src/main/java/info/blockchain/wallet/view/MainActivity.java
```diff
@@ -186,12 +186,16 @@ private void startScanActivity() {
...
-        args.putString("scan_data", strResult);
-        fragment.setArguments(args);
-        FragmentManager fragmentManager = getFragmentManager();
-        fragmentManager.beginTransaction().replace(R.id.content_frame, fragment).commit();
+//        Fragment fragment = new SendActivity();
+//        Bundle args = new Bundle();
+//        args.putString("scan_data", strResult);
+//        fragment.setArguments(args);
+//        FragmentManager fragmentManager = getFragmentManager();
+//        fragmentManager.beginTransaction().replace(R.id.content_frame, fragment).commit();
```
mpfluger: Please remove commented out code or mention why it's commented out

PR #290, opened 2016-08-12
[Trivial] Fix few french words

Kirvx:


riaanjvos:
Thank you for the pull request @Kirvx

PR #304, opened 2016-08-25
chore(Dogfood flavour) Dogfood builds for internal testing. Build too…

ditn:
- Dogfood builds for internal testing, can be run alongside normal builds
- Removed long since deprecated permissions from Manifest & unnecessary labels (these are inherited from the Application class anyway)
- Updated build tools, target SDK 24
- Add "Thunk" annotation
- Fix memory leak in SendActivity
- Moved screenshot disabling to BaseAuthActivity


app/src/main/java/info/blockchain/wallet/view/SendActivity.java
```diff
@@ -640,6 +621,8 @@ public void onShowPaymentDetails(PaymentConfirmationDetails details) {
         });
 
         alertDialog.show();
+        // To prevent the dialog from appearing too large on Android N
+        alertDialog.getWindow().setLayout(WindowManager.LayoutParams.MATCH_PARENT, WindowManager.LayoutParams.WRAP_CONTENT);
```
ditn: This is a weird bug. Not seen it on any other custom dialog and only on Android N.

app/src/main/java/info/blockchain/wallet/view/SendActivity.java
```diff
@@ -288,24 +287,6 @@ private void setupDestinationView(){
...
-            @Override
-            public void onGlobalLayout() {
-                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN) {
-                    binding.accounts.spinner.getViewTreeObserver().removeOnGlobalLayoutListener(this);
-                } else {
-                    binding.accounts.spinner.getViewTreeObserver().removeGlobalOnLayoutListener(this);
-                }
-
-                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN) {
-                    binding.accounts.spinner.setDropDownWidth(binding.accounts.spinner.getWidth());
```
ditn: This doesn't work - it sets the width of the view to the current width!

riaanjvos: Just tested and this does still work. I remember we added this to allow a small gap on the left hand side of the drop down so the user can click on screen to hide the drop down again. This does however depend on screen size and SDK version.

PR #306, opened 2016-08-30
Feature/password reveal warning

ditn:
- Warning dialog for passwords exposed by InputTextLayout
- Fancy new vector-based animation for the reveal icon


kristovatlas:
Nice, concept ACK

PR #307, opened 2016-08-31
feat(Sms notification): Added toggle to enable notification via sms.

riaanjvos:


ditn:
👍

app/src/main/res/values/strings.xml
```diff
@@ -580,4 +580,8 @@
     <string name="accept_higher_fee">Accept higher fee</string>
     <string name="large_tx_high_fee_warning">"This transaction requires a higher fee for dust consumption due to the small amount of change to be returned."</string>
     <string name="password_reveal_warning">Be careful! Revealing this text allows your password to be copied to the clipboard, which can then be read by all applications on your device. Also be aware of people standing behind you!</string>
+
+    <string name="sms_notifications">Sms Notifications</string>
```
ditn: Super minor, but Sms should probably be all caps

riaanjvos: Good catch. That definitely needs to be caps. Thanks

PR #320, opened 2016-09-08
Backend environment

riaanjvos:
- All web calls have been refactored out into the My-Wallet-v3-jar
- Added backend environment settings button on landing page (Dogfood build only) to select between production, dev, staging environment.


ditn:
👍

PR #370, opened 2016-10-11
Feature/fingerprint authentication

ditn:
- Added fingerprint login for users who have appropriate devices
- Added a new nav drawer header


app/src/main/java/piuk/blockchain/android/util/PersistentPrefs.java
```diff
@@ -25,6 +25,8 @@
     String KEY_SECURITY_TIME_ELAPSED = "security_time_elapsed";
     String KEY_SECURITY_TWO_FA_NEVER = "security_two_fa_never";
     String KEY_SECURITY_BACKUP_NEVER = "security_backup_never";
+    String KEY_ENCRYPTED_PIN_CODE = "encrypted_pin_code";
```
riaanjvos: Just make 100% sure that this does not persist after use as it's obfuscated and not encrypted.

ditn: So it'll persist until it's cleared - either by turning off fingerprint login or after too many failed attempts. It is fully encrypted though - AES256

riaanjvos: Ok, great.

PR #371, opened 2016-10-12
Reduce warnings

riaanjvos:


app/src/main/java/piuk/blockchain/android/data/websocket/WebSocketHandler.java
```diff
@@ -266,10 +266,8 @@ public void onTextMessage(WebSocket websocket, String message) {
                                                         } else if (in_addr == null) {
                                                             in_addr = (String) prevOutObj.get("addr");
                                                         } else {
-                                                            ;
```
ditn: Might as well get rid of these empty catch blocks if we're not doing anything with them

riaanjvos: Good catch. I'll get rid of them.

PR #404, opened 2016-11-22
Bug fixes/regression 16 11 2016

ditn:


riaanjvos:
Some tests in the HD-jar has failed. This will need to be addressed before this PR can be merged.

PR #451, opened 2017-03-08
Fix/lint fixes

ditn:
Mostly lint error fixes. Removed some unused resources and fixed a bug where the Settings page would load constantly.


kristovatlas:
Thanks for linting!

PR #456, opened 2017-03-21
Fix/regression 6 3 5

riaanjvos:


app/src/main/java/piuk/blockchain/android/ui/account/AccountViewModel.java
```diff
@@ -229,7 +230,10 @@ void confirmImportWatchOnly(String address) {
         legacyAddress.setCreatedDeviceName("android");
         legacyAddress.setCreatedTime(System.currentTimeMillis());
         legacyAddress.setCreatedDeviceVersion(BuildConfig.VERSION_NAME);
-        dataListener.showRenameImportedAddressDialog(legacyAddress);
+
+        compositeDisposable.add(accountDataManager.updateLegacyAddress(legacyAddress).subscribe(() -> {
+            dataListener.showRenameImportedAddressDialog(legacyAddress);
+        }));
```
ditn: There's no error handling here, could throw `OnErrorNotImplementedException` and crash the app if the sync fails

riaanjvos: Good catch, thanks. Updated.

PR #472, opened 2017-04-05
feat(Buy): Add get bitcoin link to empty transactions screen.

riaanjvos:
This will add a 'Get Bitcoin' button on the empty transactions screen.
SEPA user will be directed to buy bitcoin.
Non-SEPA users will be directed to receive bitcoin via QR.

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -681,6 +681,7 @@
 
     <!-- BUY BITCOIN -->
     <string name="buy_bitcoin">Buy Bitcoin</string>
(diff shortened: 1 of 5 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -681,6 +681,7 @@
 
     <!-- BUY BITCOIN -->
     <string name="buy_bitcoin">Buy Bitcoin</string>
+    <string name="get_bitcoin">GET BITCOIN</string>
```
ditn: This should be lowercase with the first letter caps - the text will be capitalised on the button automatically. No big deal though.

PR #476, opened 2017-04-06
Feature/ui fixes

ditn:
Mostly an update to use cards as per the designs, fade in complete view, allow resetting of onboarding flag in prefs via debug settings.


app/src/main/res/layout/include_no_transaction_message.xml
```diff
@@ -13,6 +13,7 @@
         tools:visibility="visible">
 
         <LinearLayout
+            android:animateLayoutChanges="true"
```
riaanjvos: Nice haha

PR #477, opened 2017-04-06
fix(Calligraphy): Fix calligraphy related crash on older devices.

riaanjvos:


riaanjvos:
Yeah. Tested on the Galaxy S3 mini.

PR #478, opened 2017-04-06
fix(Onboarding): Remove end-state once user leaves overview screen.

riaanjvos:


app/src/main/java/piuk/blockchain/android/ui/balance/BalanceFragment.java
```diff
@@ -639,6 +639,7 @@ private void initOnboardingPager() {
                 public void onPageScrolled(int position, float positionOffset, int positionOffsetPixels) {
                     if (position == viewModel.getOnboardingPages().size()) {
                         binding.noTransactionInclude.onboardingViewpagerLayout.pagerOnboarding.setPagingEnabled(false);
+                        viewModel.setOnboardingComplete(true);
```
ditn: Woops, my bad.

PR #506, opened 2017-05-02
Buy

jtormey:


app/src/main/java/piuk/blockchain/android/data/services/ExchangeService.java
```diff
@@ -0,0 +1,123 @@
...
+                    metadataSubject.onNext(exchangeData);
+                } catch (Exception e) {
+                    Log.d(TAG, "reloadExchangeMetadata error: " + e.getMessage());
+                }
+
+            }
+        }.start();
+    }
+
+    private Metadata getMetadata() throws Exception {
```
ditn: If this is wrapped in Rx, you won't need to manually create a new thread for it. We try to avoid creating anonymous threads in Android as much as possible

app/src/main/java/piuk/blockchain/android/data/services/ExchangeService.java
```diff
@@ -0,0 +1,123 @@
...
+    public static final String TAG = ExchangeService.class.getSimpleName();
+
+    private PayloadManager payloadManager;
+    private ReplaySubject<Metadata> metadataSubject;
+    private boolean didStartLoad;
+
+    @Inject PayloadDataManager payloadDataManager;
+
+    private ExchangeService() {
+        this.payloadManager = PayloadManager.getInstance();
```
ditn: I'd pass the `PayloadManager` through the constructor here. If we can avoid using `@Inject` annotations except at the top-most level (ie in Presenters/ViewModels), it makes testing a lot easier and is less brittle going forwards

app/src/main/java/piuk/blockchain/android/ui/buy/BuyActivity.java
```diff
@@ -52,14 +64,17 @@ protected void onCreate(Bundle savedInstanceState) {
         coinifyApi = new CoinifyApi();
 
         WebView webView = binding.webview;
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
```
ditn: Minor niggle, but we have an `AndroidUtils` class which does these comparisons and makes it easy to spot sections which are API dependent later on

app/src/main/java/piuk/blockchain/android/ui/buy/BuyActivity.java
```diff
@@ -167,4 +162,22 @@ public void fetchTrades(String accessToken) {
...
+    private void showTradeComplete(long date, String receiveAddress, String txHash) {
+        String tradeDetailsMessage = getString(R.string.trade_complete_details);
+        // TODO: format date
+        String alertMessage = String.format(Locale.getDefault(), tradeDetailsMessage, "{date}");
+
+        new AlertDialog.Builder(this, R.style.AlertDialogStyle)
+                .setTitle(getString(R.string.trade_complete))
+                .setMessage(alertMessage)
+                .setCancelable(false)
+                .setPositiveButton(R.string.ok_cap, (dialog, whichButton) -> {
```
ditn: This can be simplified to `setPositiveButton(R.string.ok_cap, null)` for a one-liner

app/src/main/java/piuk/blockchain/android/ui/buy/FrontendJavascriptManager.java
```diff
@@ -36,8 +36,8 @@ public void activateMobileBuy(String uid, String sharedKey, String password) {
         executeScript(script);
     }
 
-    public void activateMobileBuyFromJson(String walletJson, String externalJson, String magicHash, String password) {
-        String script = FrontendJavascriptManager.createActivateFromJsonScript(walletJson, externalJson, magicHash, password);
+    public void activateMobileBuyFromJson(String walletJson, String externalJson, String magicHash, String password, Boolean firstLogin) {
```
ditn: Is there any reason for using `Boolean` rather than `boolean` here?

jtormey: Nope, will change to a literal.

app/src/main/java/piuk/blockchain/android/data/services/ExchangeService.java
jtormey: Do you mean `PayloadDataManager`? And if I pass it to through the constructor, do I then need to pass it through each call to `getInstance`? (Even though it may already have an instance, and not need the PayloadDataManager again).

ditn: `PayloadManager` at this level. Well that's the thing, I'm not sure why this is a Singleton, and if it needs to be a Singleton, have Dagger create it as such for you. If not, just `new` it into your DataManager. I'm guessing you want it as a single instance because of the `ReplaySubject`?

jtormey: Yes, I made it a singleton because I'd rather not reload the exchange metadata every time the view loads. With ReplaySubject, observers can get the latest data but also be notified it changes. Is there a better pattern I could use for that?

Wasn't aware of creating singletons with Dagger, seems like a much better idea than what I'm doing haha. Although, maybe it would be better to move the functionality of ExchangeService back into BuyDataManager and just have that be the singleton?

ditn: Possibly, yeah. If you move much of that functionality into a DataManager and then tell Dagger how to instantiate that in the `ApiModule`, it'll be provided as a Singleton. I've already done that for the ContactsDataManager. 

Going forwards I want to create persistent repos for the `Settings`, `MetaData` and a few other objects but I've not got around to it yet

jtormey: Ok cool, will give that a try, thanks.

app/src/main/java/piuk/blockchain/android/data/datamanagers/BuyDataManager.java
```diff
@@ -0,0 +1,55 @@
...
+                (isSepa, isInvited) -> isSepa && isInvited
+        );
+    }
+
+    private Observable<Boolean> getIsInvited() {
+        return this.settingsDataManager.initSettings(
+                payloadDataManager.getWallet().getGuid(),
+                payloadDataManager.getWallet().getSharedKey()
+        ).map(settings -> {
+            // TODO: implement settings.invited.sfox
```
ditn: Is this unneeded for now?

jtormey: It's needed for QA to test SFOX, but since SFOX is on hold currently I'm not sure it's worth it.

ditn: Ahh cool.

app/src/main/java/piuk/blockchain/android/ui/buy/FrontendJavascriptManager.java
```diff
@@ -48,25 +63,27 @@ public void teardown() {
     private void executeScript(String script) {
         Log.d(TAG, "Executing: " + script);
         // TODO: 06/04/2017 evaluateJavascript isn't available on pre-19 devices
-        webView.post(() -> webView.evaluateJavascript(script, frontendJavascript));
+        new Handler(Looper.getMainLooper()).post(() -> webView.evaluateJavascript(script, frontendJavascript));
```
ditn: So this is an issue - buy/sell will completely explode on older devices.

Either we need to be using a library like [js-evaluator](https://github.com/evgenyneu/js-evaluator-for-android) or we need to disable/hide buy/sell in devices pre-19 (which is my preferred option).

app/src/main/java/piuk/blockchain/android/ui/buy/FrontendJavascriptManager.java
```diff
@@ -48,25 +63,27 @@ public void teardown() {
     private void executeScript(String script) {
         Log.d(TAG, "Executing: " + script);
```
riaanjvos: Might be best to remove this if anything sensitive is logged here. Not too sure but createCheckForCompletedTradesScript() seems to contain wallet json data?

app/src/main/java/piuk/blockchain/android/ui/home/MainViewModel.java
```diff
@@ -373,10 +380,14 @@ private void preLaunchChecks() {
                                 }
                             })
                             .subscribe(
-                                    isSepaCountry -> {
-                                        if (isSepaCountry && BuildConfig.BUY_BITCOIN_ENABLED) {
-                                            enableBuySell();
-                                        }
+                                    canBuy -> {
+                                        Log.d(TAG, "preLaunchChecks: canBuy " + canBuy);
+                                        dataListener.setBuySellEnabled(canBuy && BuildConfig.BUY_BITCOIN_ENABLED);
```
ditn: Ideally we'd pass this through an interface from the UI as otherwise it could be tricky to test based on gradle config. The other advantage is that we can return `BuildConfig.BUY_BITCOIN_ENABLED && AndroidUtils.is19orHigher()` whilst keeping Android-specific code out of the UI.

app/src/main/java/piuk/blockchain/android/ui/home/MainActivity.java
```diff
@@ -727,8 +758,52 @@ private void setupBuyWebView() {
         // Setup buy WebView
         // TODO: 17/03/2017 Check if there's a better way to improve loading time of this webview
         buyWebView = new WebView(this);
+        if (AndroidUtils.is21orHigher()) {
+            buyWebView.setWebContentsDebuggingEnabled(BuildConfig.DEBUG);
+        }
         buyWebView.getSettings().setJavaScriptEnabled(true);
         buyWebView.loadUrl("http://localhost:8080/wallet/#/intermediate");
```
ditn: Should this not be changed?

riaanjvos: Will QA have a local environment for this?

jtormey: Yeah, would be best to have it stored in the environment. On iOS, it can be changed in the debug menu for QA / development.

app/src/main/java/piuk/blockchain/android/ui/buy/FrontendJavascriptManager.java
jtormey: Sure, should be simple to do that.

app/src/main/java/piuk/blockchain/android/ui/buy/FrontendJavascriptManager.java
jtormey: Changed so that it only logs when debugging, is that fine?

ditn: That's great, nice one.

PR #510, opened 2017-05-02
Move ExchangeService functionality into BuyDataManager

jtormey:
@ditn following the discussion from #506, I've moved ExchangedService into BuyDataManager, having both does not seem necessary. I wasn't able to figure out how to get Dagger to do Singleton DI though, could you help with that? (I know the `getInstance` solution in this PR is wrong).


ditn:
Ah yeah I can see why you're having trouble with this now. I'll take a look at this later and see what I can do, might need to rejig dagger a bit.

PR #511, opened 2017-05-03
Fix/tx feed update

riaanjvos:


app/src/main/java/piuk/blockchain/android/ui/home/MainActivity.java
```diff
@@ -378,7 +386,7 @@ public void selectDrawerItem(MenuItem menuItem) {
                 startActivityForResult(new Intent(MainActivity.this, BackupWalletActivity.class), REQUEST_BACKUP);
                 break;
             case R.id.nav_addresses:
-                startActivity(new Intent(MainActivity.this, AccountActivity.class));
+                startActivityForResult(new Intent(MainActivity.this, AccountActivity.class), ACCOUNT_EDIT);
```
ditn: Wait, what result are we getting from this?

PR #531, opened 2017-05-22
fix(Send): Invalidate receiving address after user edits field.

riaanjvos:
Added a on text change listener to invalidate receive address kept in memory after scanning a receive address. Some devices ignore the on click listener that was put in place to prevent this behaviour.

### DIFF
--- a/app/src/main/java/piuk/blockchain/android/ui/send/SendFragment.java
+++ b/app/src/main/java/piuk/blockchain/android/ui/send/SendFragment.java
@@ -29,6 +29,7 @@ import android.text.InputType;
 import android.text.TextWatcher;
 import android.text.method.DigitsKeyListener;
 import android.util.Log;
+import android.view.KeyEvent;
 import android.view.LayoutInflater;
 import android.view.Menu;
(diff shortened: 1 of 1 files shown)


app/src/main/java/piuk/blockchain/android/ui/send/SendFragment.java
```diff
@@ -460,6 +462,25 @@ private void setupDestinationView() {
             v.performClick();
             return false;
         });
+        binding.destination.addTextChangedListener(new TextWatcher() {
```
ditn: If you're feeling fancy you can swap to:

```
RxTextView.textChanges(binding.destination)
                .doOnNext(ignored -> viewModel.setReceivingAddress(null))
                .subscribe(new IgnorableDefaultObserver());```

But it totally doesn't matter - just an option we have now!

riaanjvos: I did not know that.
I am feeling fancy and this is much neater. I will change the PR shortly.

PR #541, opened 2017-06-06
Feature/web pair

riaanjvos:


app/src/main/java/piuk/blockchain/android/ui/pairing_code/PairingCodePresenter.java
```diff
@@ -0,0 +1,69 @@
...
+        //no op
+    }
+
+    public String getFirstStep() {
+        return String.format(stringUtils.getString(R.string.pairing_code_instruction_1), webWalletUrl);
+    }
+
+    public void generatePairingQr() {
+
+        getCompositeDisposable().add(getPairingEncryptionPasswordObservable()
```
ditn: It doesn't matter at all, but if you want to you can call `.compose(RxUtil.addObservableToCompositeDisposable(this))` just before subscribing in the chain to achieve the same thing.

app/src/main/java/piuk/blockchain/android/ui/pairing_code/PairingCodeView.kt
```diff
@@ -0,0 +1,18 @@
...
+import android.support.annotation.StringRes
+
+import piuk.blockchain.android.ui.base.View
+import piuk.blockchain.android.ui.customviews.ToastCustom
+
+interface PairingCodeView : View {
+
+    fun onQrLoaded(bitmap: Bitmap)
+
+    fun showToast(@StringRes message: Int, @ToastCustom.ToastType toastType: String)
```
ditn: Looks weird with the annotations on one side and the type on the other, silly Kotlin ¯\_(ツ)_/¯

riaanjvos: Ha! Actually didn't mean to leave this class in Kotlin. But no harm caused by leaving it as is.

ditn: dat Java interop

PR #545, opened 2017-06-07
Fix/onboarding

ditn:


riaanjvos:
Some jar conflicts need fixing before merging

PR #549, opened 2017-06-08
Feat/buy

riaanjvos:
Calculate if mobile buy should be enabled for wallet based on user guid. Add performance warning.


app/src/main/java/piuk/blockchain/android/data/datamanagers/BuyDataManager.java
```diff
@@ -34,10 +35,16 @@ public BuyDataManager(OnboardingDataManager onboardingDataManager,
     }
 
     public Observable<Boolean> getCanBuy() {
+
+        if(BuildConfig.DEBUG) {
+            return Observable.just(true);
```
ditn: Haha nice

PR #550, opened 2017-06-08
Development

riaanjvos:


app/src/main/java/piuk/blockchain/android/ui/customviews/AnnouncementView.java
```diff
@@ -0,0 +1,73 @@
...
+
+        ivClose = (ImageView)rl.getChildAt(0);
+        ivImage = (ImageView)rl.getChildAt(1);
+        LinearLayout llmain = (LinearLayout)rl.getChildAt(2);
+        LinearLayout ll = (LinearLayout)llmain.getChildAt(0);
+        tvContent = (TextView)llmain.getChildAt(1);
+        tvLink = (TextView)rl.getChildAt(3);
+
+        tvTitle = (TextView)ll.getChildAt(0);
+        ivEmoji = (ImageView)ll.getChildAt(1);
```
ditn: Why are we getting these as children rather than using `findViewById(...)`?

riaanjvos: Yeah that would have been the better route. I'll update and make a new PR for this.

ditn: Nah don't worry, if it works it's fine. Can be updated some other day. Just wondered if there was a particular reason for it.

PR #553, opened 2017-06-09
Fix/onboarding pages

riaanjvos:
This will allow mobile buy to be visible for all environments other than production, based on the build config flag - BUY_BITCOIN_ENABLED

Fixed race condition where ticker would sometimes not complete in time for onboarding pager to show correct exchange rate. Balance needs a refactor to make it less brittle.


app/src/main/java/piuk/blockchain/android/data/datamanagers/BuyDataManager.java
```diff
@@ -34,17 +34,17 @@ public BuyDataManager(OnboardingDataManager onboardingDataManager,
         return exchangeService.watchPendingTrades();
     }
 
-    public Observable<Boolean> getCanBuy() {
+    public synchronized Observable<Boolean> getCanBuy() {
 
-        if(BuildConfig.DEBUG) {
+        if (!PersistentUrls.getInstance().getCurrentEnvironment().equals(PersistentUrls.Environment.PRODUCTION)) {
```
ditn: Oooh this will break because of the refactor I just did, inject `EnvironmentSettings` and then call `environmentSettings.getEnvironment().equals(...)`

app/src/main/java/piuk/blockchain/android/ui/balance/BalanceViewModel.java
```diff
@@ -792,27 +806,29 @@ public void setOnboardingComplete(boolean completed) {
...
-            dataListener.startBuyActivity();
-        } else {
-            dataListener.startReceiveFragment();
-        }
+        compositeDisposable.add(buyDataManager.getCanBuy().subscribe(buyAllowed -> {
+            if (buyAllowed) {
+                dataListener.startBuyActivity();
+            } else {
+                dataListener.startReceiveFragment();
+            }
```
ditn: Error not implemented

PR #557, opened 2017-06-12
Refresh magic hash when loading exchange data

jtormey:
_Problem_

After the web wallet saves data to the metadata service, if the web view is closed and reopened, Android will fetch the new metadata, but will not get the new magic hash. Android then passes an incorrect magic has to the web wallet, causing saves to metadata to not work.

_Solution_

Fetch the latest magic hash before passing it to the web wallet.

I'm not sure this is the best way to do this, I think it might be better to fix this in v3-jar. I tried making it so the magic hash will update automatically during calls to Metadata::getMetadata, but couldn't get that to work. Let me know if you have any ideas @ditn @riaanjvos

### DIFF
--- a/app/src/main/java/piuk/blockchain/android/data/services/ExchangeService.java
+++ b/app/src/main/java/piuk/blockchain/android/data/services/ExchangeService.java
@@ -75,10 +75,14 @@ public class ExchangeService {
                         })
                         .compose(RxUtil.applySchedulersToObservable())
                 ),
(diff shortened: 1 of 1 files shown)


riaanjvos:
I see your predicament. The magic hash gets messy as soon as we start using the same wallet cross platform simultaneously.
Jar has been updated so you don't have to fetch magic hash manually anymore.
https://github.com/blockchain/My-Wallet-V3-jar/pull/136

PR #559, opened 2017-06-13
Disable buy when second password is active

jtormey:
Buy needs access to the metadata service, so it can't be used when second password is active. This PR makes it so that an alert shows, instead of a never ending spinner.

On web and iOS, we get around the second password restriction by using the metadataHDNode property that is set on the payload. That's definitely the better solution, but wasn't sure of the best way to use that here. Any ideas on how to do that? If it's difficult, this alert should be fine in the meantime.


ditn:
Interesting, I thought we'd already done this. We'll take a look tomorrow.

Adam Bennett | 07855 230 708
Android Developer at Blockchain

app/src/main/res/values/strings.xml
```diff
@@ -699,6 +699,8 @@
     <string name="onboarding_available_now">Available now</string>
     <string name="onboarding_buy_details">Buy bitcoin directly from your mobile wallet. It\'s simple, seamless, and secure.</string>
     <string name="onboarding_buy_performance_warning">The buy experience may be a little slow on your device. Thanks in advance for your patience.</string>
+    <string name="buy_second_password_alert_title">Buy Unavailable</string>
```
ditn: Please try to group these with other buy strings if possible

ditn:
So we've definitely done this for Contacts but not buy, I'll look at porting across the Contacts method today.

riaanjvos:
Buy will be hidden for now if second password is enabled.
We'll get past the second password issue by implementing the same solution as Contacts (metadataHDNode )

PR #564, opened 2017-06-14
Prevent redirecting to browser when buy is closed

jtormey:
This fixes the issue where pressing the "back" button will open blockchain.info in the phone's browser, on some phones. I think what was happening is that the teardown() function was making a redirect, which is sometimes intercepted by Android. Teardown() is really just to refresh the page, this changes it so that the web view is refreshed directly, preventing any redirect.


ditn:
Interesting, the alternative is to capture any back button inputs and call `finish()` if the webview is at a certain point. This seems like a fine solution though.

PR #565, opened 2017-06-15
Gradle Update

ditn:
Update Gradle to 4.0 for much faster builds, update various dependencies incl. Google Services. Change Travis script to only build one version of the app 🙌


ditn:
Okay this should actually be ready to merge now.

PR #568, opened 2017-06-19
Fix - Buy trades

jtormey:
There was an issue with mapping exchange data JSON to java objects. I couldn't reproduce on any of the phones I tried, but from what I've read it seems like the issue could be missing default constructors. Would that make sense? Error message in the ticket contains:

```java
06-14 13:51:26.216 8012-8012/? W/System.err: com.fasterxml.jackson.databind.JsonMappingException: No suitable constructor found for type [simple type, class piuk.blockchain.android.data.exchange.CoinifyData]: can not instantiate from JSON object (missing default constructor or creator, or perhaps need to add/enable type information?)
```

I'm also filtering out sell trades now, since those don't need to be watched.


ditn:
>Would that make sense?

Certainly seems that way, we've seen it a few times before with other classes. I'm surprised it's difficult to reproduce though.

PR #571, opened 2017-06-22
Fix/currency unit restore

riaanjvos:


riaanjvos:
Based this off of the refactor/balance branch accidentally - We'll have to first merge balance, sorry

PR #574, opened 2017-06-23
refactor(Account): Remove redundant code.

riaanjvos:


app/src/main/java/piuk/blockchain/android/ui/account/AccountEditViewModel.java
```diff
@@ -160,52 +160,41 @@ public void onViewReady() {
...
-                List<LegacyAddress> legacy = iAccount.getLegacyAddresses();
-                legacyAddress = legacy.get(addressIndex);
-
-                accountModel.setLabel(legacyAddress.getLabel());
-                accountModel.setLabelHeader(stringUtils.getString(R.string.name));
-                accountModel.setXpubDescriptionVisibility(View.GONE);
-                accountModel.setXpubText(stringUtils.getString(R.string.address));
-                accountModel.setDefaultAccountVisibility(View.GONE);//No default for V2
-                setArchive(legacyAddress.getTag() == LegacyAddress.ARCHIVED_ADDRESS);
+            if (payloadDataManager.getWallet().isUpgraded()) {
```
ditn: Is any of the stuff in this block even necessary though? In theory users on V2 shouldn't be able to get to this point

riaanjvos: Yeah it's till needed in this case. Looks like the "V2" just refers to a legacy address being edited.

ditn: Right, but we're also checking that the wallet is upgraded? That should always be true, right?

riaanjvos: Correct, from now on it will always be true because we force a V2 wallet to be upgraded to V3. There are a few places in the wallet that this could be removed now, we could do all these in one go to clean up the code a bit (maybe a separate PR)

ditn: Fair. I've been removing them as I see them but we should probably do it that way.

PR #584, opened 2017-06-29
fix(Pairing): Increase web pairing QR code dimensions.

riaanjvos:


riaanjvos:
True true. Updating this now.

PR #606, opened 2017-07-11
Fix/minor bugs 06 2017

riaanjvos:


build.gradle
```diff
@@ -17,7 +17,7 @@ buildscript {
         google()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.3.2'
+        classpath 'com.android.tools.build:gradle:3.0.0-alpha6'
```
ditn: Hmmm this is breaking CI at the moment and we probably shouldn't be using it until it's finalised. Best to roll this back. Otherwise these are solid changes.

PR #607, opened 2017-07-11
Fix/jira fixes

ditn:


app/src/main/res/values/strings.xml
```diff
@@ -432,7 +433,7 @@
     <string name="new_password">New Password</string>
     <string name="sms_verified">Mobile number verified</string>
     <string name="preferences">Preferences</string>
-    <string name="two_fa_summary">Protect your wallet from unauthorized access by enabling 2-Step Verification. You can choose to use your mobile phone number to secure your wallet. This code will be requested when you try to access your wallet via the web browser.\n\nIn order to use other authentication methods instead, please login to our web wallet.</string>
+    <string name="two_fa_summary">Protect your wallet from unauthorized access by enabling 2-Step Verification. You can choose to use your mobile phone number to secure your wallet. This code will be requested when you try to access your wallet via the web browser.\n\nIn order to use other authentication methods instead, please login to our web wallet at <a href="https://blockchain.info/wallet/login/">blockchain.info/wallet/login</a></string>
```
ditn: Replacement strings? As in just define a new string name?

riaanjvos: String format arguments using %1$s

PR #644, opened 2017-08-03
Feature/contacts hide

riaanjvos:


app/src/main/java/piuk/blockchain/android/ui/chooser/AccountChooserPresenter.java
```diff
@@ -93,17 +96,23 @@ private void loadContactsOnly() {
...
-        return contactsDataManager.getContactList()
-                .filter(ContactsPredicates.filterByConfirmed())
-                .toList()
-                .doOnSuccess(contacts -> {
-                    if (!contacts.isEmpty()) {
-                        itemAccounts.add(new ItemAccount(stringUtils.getString(R.string.contacts_title), null, null, null, null));
-                        for (Contact contact : contacts) {
-                            itemAccounts.add(new ItemAccount(null, null, null, null, contact, null));
+
+        if (!BuildConfig.CONTACTS_ENABLED) {
```
ditn: Access this via the view interface, otherwise tests will fail when we change the value

PR #658, opened 2017-08-08
Development

riaanjvos:


app/build.gradle
```diff
@@ -21,10 +21,8 @@ android {
         testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"
         vectorDrawables.useSupportLibrary = true
         archivesBaseName = "blockchain-$versionName"
-
-        // Flag for Buy and Contacts feature
-        buildConfigField "boolean", "CONTACTS_ENABLED", "false"
-        buildConfigField "boolean", "BUY_BITCOIN_ENABLED", "true"
+        // Flag to hide or show the Contacts feature
+        buildConfigField "boolean", "CONTACTS_ENABLED", "true"
```
ditn: We should probably leave this disabled on `master` as that's what's actually going out, but we can do that when we increment the version codes.

riaanjvos: True, will update with version codes. Accidentally committed this as true.

PR #664, opened 2017-08-11
fix(Onboarding): Fix issue where onboarding could display again if th…

ditn:
…e user returned to the home screen after dismissing the onboarding pager


ditn:
Okay I'm going to stop committing to this PR now 😂

PR #685, opened 2017-08-30
Feature/eth exchange rates

ditn:


ditn:
* Exchange rates added
* Loaded actual ETH account for balance
* Updated a few bits and pieces

Balance page is pretty slow right now and some stuff needs caching, but I figure this can be fixed once the loading is simplified when all the announcement stuff is moved out to the dashboard.

PR #691, opened 2017-09-05
Feature/ethereum send

riaanjvos:


app/src/main/java/piuk/blockchain/android/data/cache/DynamicFeeCache.java
```diff
@@ -3,25 +3,37 @@
...
+    }
+
+    public void setEthFeeOptions(FeeOptions ethFeeOptions) {
+        this.ethFeeOptions = ethFeeOptions;
     }
 
     public void destroy() {
-        feeOptions = null;
+        btcFeeOptions = null;
+        ethFeeOptions = null;
```
ditn: I don't know why we clear this cache actually, doesn't really seem necessary 🤔

app/src/main/java/piuk/blockchain/android/ui/receive/ReceiveFragment.kt
```diff
@@ -144,22 +144,22 @@ class ReceiveFragment : BaseFragment<ReceiveView, ReceivePresenter>(), ReceiveVi
...
             hint = "0${defaultDecimalSeparator}00"
             addTextChangedListener(fiatTextWatcher)
             disableSoftKeyboard()
         }
 
         // Units
-        currency_btc.text = presenter.currencyHelper.btcUnit
-        currency_fiat.text = presenter.currencyHelper.fiatUnit
+        currencyCrypto.text = presenter.currencyHelper.btcUnit
+        currencyFiat.text = presenter.currencyHelper.fiatUnit
```
ditn: Stick to snake_case if possible, this will be fixed in future versions of the Kotlin plugin

app/src/main/java/piuk/blockchain/android/data/cache/DynamicFeeCache.java
riaanjvos: I was just following the btc convention. destroy() literally doesn't even get used, I'll get rid of it

ditn: Yeah it's cool, I'm just not sure why it was there

app/src/main/java/piuk/blockchain/android/ui/receive/ReceiveFragment.kt
riaanjvos: I would honestly prefer camel case in the code, but I don't mind, as long as we stick to one of them

PR #696, opened 2017-09-13
Feature/ethereum send

riaanjvos:
Just a PR to sync up


app/src/main/java/piuk/blockchain/android/ui/receive/ReceiveCurrencyHelper.java
```diff
@@ -224,4 +228,54 @@ public String getTextFromSatoshis(long satoshis, String decimalSeparator) {
...
+        String amountToSend = stripSeparator(text, decimalSeparator);
+
+        Double amount;
+        try {
+            amount = java.lang.Double.parseDouble(amountToSend);
+        } catch (NumberFormatException e) {
+            amount = 0.0;
+        }
+
+        long amountL = Convert.toWei(amount.toString(), Convert.Unit.ETHER).longValue();
```
ditn: Here you should probably do `Convert.toWei(amount.toString(), Convert.Unit.ETHER).toBigInteger()`, otherwise by switching to `long` you could end up with an overflow.

app/src/main/java/piuk/blockchain/android/ui/send/PendingTransaction.java
```diff
@@ -21,7 +22,61 @@
...
+
+    @JsonIgnore
+    public boolean isWatchOnly() {
+
+        boolean watchOnly = false;
+
+        if(!isHD()) {
+            if(sendingObject.getAccountObject() instanceof LegacyAddress) {
+                LegacyAddress legacyAddress = (LegacyAddress)sendingObject.getAccountObject();
+                watchOnly = legacyAddress.isWatchOnly() && (legacyAddress.getPrivateKey() == null ||  legacyAddress.getPrivateKey().isEmpty());
```
ditn: Isn't much of this unnecessary? `isWatchOnly` checks for the privkey anyway

app/src/main/java/piuk/blockchain/android/ui/receive/ReceiveCurrencyHelper.java
riaanjvos: Good catch. Didn't even see this.

app/src/main/java/piuk/blockchain/android/ui/send/PendingTransaction.java
riaanjvos: I think this is for a watch only wallet where the private key has already been scanned for the pending transaction. Will double check.

PR #702, opened 2017-09-19
Feature/ethereum send

riaanjvos:


app/src/main/java/piuk/blockchain/android/data/ethereum/EthDataManager.kt
```diff
@@ -137,17 +140,39 @@ class EthDataManager(
                 .compose(RxUtil.applySchedulersToObservable())
     }
 
+    //TODO we need to derive private key if we're going to spend eth - no point in just using address from metadata unless double encryted
```
ditn: Yeah this is something I don't understand about why we're using metadata to store the address...

mpfluger: Otherwise we'd have to ask 2nd pw users on every login just to be able to derive the address. For send we have to ask, but that's the point of second password.

ditn: Gotcha. Makes sense.

riaanjvos: Yeah I saw this afterwards. Fix in progress 👍

PR #732, opened 2017-10-04
Fix/crypto currency tab

riaanjvos:


app/src/main/java/piuk/blockchain/android/ui/dashboard/DashboardPresenter.kt
```diff
@@ -75,16 +76,26 @@ class DashboardPresenter @Inject constructor(
...
     }
 
     internal fun onResume() {
+        cryptoCurrency = currencyState.cryptoCurrency
         updateChartsData(timeSpan)
         updateAllBalances()
         updatePrices()
     }
 
+    fun getCurrentCryptoCurrency(): Int {
```
mpfluger: I don't get this - where is it used and why do we need to go from the enum to ints?

riaanjvos: It's for the tab layout that only accepts ints :)

PR #736, opened 2017-10-16
fix(Ether): Fix unconfirmed tx check. Fix rounding issue.

riaanjvos:


app/src/main/java/piuk/blockchain/android/data/ethereum/EthDataManager.kt
```diff
@@ -84,48 +86,22 @@ class EthDataManager(
...
-            }
+    fun hasUnconfirmedEthTransactions(): Observable<Boolean> {
+
+        val lastTxHash = ethDataStore.ethWallet?.lastTransactionHash
+
+        return Observable.fromIterable(ethDataStore.ethAddressResponse?.getTransactions())
+                .filter{ list -> list.hash == lastTxHash }
+                .toList()
+                .flatMapObservable { Observable.just(it.size == 0) }
+    }
```
ditn: Right, but this ignores completely if the transaction isn't yet confirmed? If there's `lastTransactionHash` that matches any Tx in the list, it'll never let you send again

ditn: You also can't guarantee that a Tx is in that list because it hasn't necessarily been fetched - chances are it hasn't. `isTxUnConfirmed` definitely still needs to be there as far as I can tell.

riaanjvos: ethDataStore.ethAddressResponse?.getTransactions() will include all confirmed transactions so we don't have to worry about checking for confirmations.
But you are right, this list might not be up to date.
I will update accordingly.

PR #749, opened 2017-10-20
fix(Send): If watch-only spend, don't attempt to fetch more keys from…

riaanjvos:
… wallet.


riaanjvos:
Double checked. It gets cleared when switching between coins and on transaction completion. On new address selected we can't clear it, but we definitely need to reset that flag. Good catch.

PR #760, opened 2017-10-31
Server Timeout

ditn:
Fix issue where uncaught exception type could cause wallets to unpair unexpectedly


ditn:
Even increased coverage. What a boss.

PR #765, opened 2017-11-06
Fix/labelling

ditn:
Fix for a couple of high priority bugs. Added logging so that we can diagnose metadata issues.


mpfluger:
One minor idea to clean it up a little bit: add functions `getLabelOrAddress` and `getLabelOrXpub`.

ditn:
Agree, I'd like to have a base interface that all of these classes inherit
from that just allow us to generically getLabel

Adam Bennett
Android Developer at Blockchain

PR #778, opened 2017-11-21
Feature/shapeshift send

ditn:


app/src/main/java/piuk/blockchain/android/data/shapeshift/ShapeShiftDataManager.kt
```diff
@@ -41,9 +83,28 @@ class ShapeShiftDataManager(private val shapeShiftApi: ShapeShiftApi, rxBus: RxB
...
+     * containing an empty list of [Trade] objects.
+     *
+     * @param masterKey The wallet master key, from [info.blockchain.wallet.bip44.HDWallet.getMasterKey]
+     * @return A [ShapeShiftTrades] object wrapping trades functionality
+     * @throws Exception Can throw various exceptions if the key is incorrect, the server is down
+     * etc
+     */
+    @WebRequest
+    @Throws(Exception::class)
+    private fun fetchOrCreateMetadataNode(masterKey: DeterministicKey): ShapeShiftTrades {
```
riaanjvos: Masterkey shouldn't be accessed like this (we'll be forced to query the user for second pw every time)
We should use payloadDataManager.metadataNodeFactory which might already contain the derived metadataHDNode.

ditn: Cool, well we can get that fixed in a bit.

PR #795, opened 2017-12-05
Fix/shapeshift tx change

riaanjvos:


app/src/main/java/piuk/blockchain/android/ui/shapeshift/models/ShapeShiftModels.kt
```diff
@@ -21,6 +21,8 @@ data class ShapeShiftData(
         var depositAmount: BigDecimal,
         // The address you're sending the funds to
         var depositAddress: String,
+        // Your change address
```
ditn: It's not really a change address though, since we're sending exact amounts. It's a refund address more than anything.

riaanjvos: It's not the refund coming back from shape shift.
It's the change coming back from the user having to split an input to get the correct amount to send.

PR #802, opened 2017-12-06
fix(ShapeShift): Add ShapeShift trade description tag to transaction …

riaanjvos:
…feed.


ditn:
`BalancePresenterTest.kt: (124, 63): No value passed for parameter 'txNoteMap'`

PR #821, opened 2018-01-15
Feat/bch data manager

riaanjvos:
We'll be able to initialise a bitcoin cash wallet from this class. As well as call any multi address / balance methods related to a bitcoin cash wallet.


app/src/main/java/piuk/blockchain/android/data/bitcoincash/BchDataManager.kt
```diff
@@ -77,30 +80,128 @@ class BchDataManager(
...
             }
         } else {
 
-            bchDataStore.bchWallet = BitcoinCashWallet.createWatchOnly(networkParameterUtils.bitcoinCashParams)
+            bchDataStore.bchWallet = BitcoinCashWallet.createWatchOnly(
+                    blockExplorer,
+                    networkParameterUtils.bitcoinCashParams)
+
+            var i = 0
+            payloadDataManager.accounts.forEach {
```
ditn: You could do an indexed foreach loop here if you wanted instead

riaanjvos: Oooo that's handy, thanks

PR #834, opened 2018-02-02
Feature/bch shapeshift support

ditn:
Mostly minor fixes and tweaks, not all relating to SS.


app/src/main/java/piuk/blockchain/android/ui/receive/ReceivePresenter.kt
```diff
@@ -116,12 +116,16 @@ class ReceivePresenter @Inject internal constructor(
...
         // Here we are assuming that the legacy address is in Base58. This may change in the future
         // if we decide to allow importing BECH32 paper wallets.
-        val address =
-                Address.fromBase58(
-                        environmentSettings.bitcoinNetworkParameters,
-                        legacyAddress.address
-                )
-        val bech32 = CashAddress.encode("bitcoincash", CashAddress.P2PKH, address.hash160)
+        val address = Address.fromBase58(
+                environmentSettings.bitcoinNetworkParameters,
```
riaanjvos: Should be environmentSettings.bitcoinCashNetworkParameters.
Technically it's the same address header and p2sh header so this wouldn't affect anything, but just more correct this way.

ditn: Hmm I thought I'd updated that... good catch

app/src/main/java/piuk/blockchain/android/ui/receive/ReceivePresenter.kt
```diff
@@ -116,12 +116,16 @@ class ReceivePresenter @Inject internal constructor(
...
+                legacyAddress.address
+        )
+        val bech32 = CashAddress.encode(
```
riaanjvos: This could just be:
val bech32 = address.toCashAddress()

app/src/main/java/piuk/blockchain/android/ui/receive/ReceivePresenter.kt
```diff
@@ -210,7 +214,11 @@ class ReceivePresenter @Inject internal constructor(
                     val address =
                             Address.fromBase58(environmentSettings.bitcoinNetworkParameters, it)
                     val bech32 =
```
riaanjvos: This could just be:
val bech32 = address.toCashAddress()

app/src/main/java/piuk/blockchain/android/ui/receive/ReceivePresenter.kt
```diff
@@ -210,7 +214,11 @@ class ReceivePresenter @Inject internal constructor(
                     val address =
                             Address.fromBase58(environmentSettings.bitcoinNetworkParameters, it)
```
riaanjvos: environmentSettings.bitcoinCashNetworkParameters

app/src/main/java/piuk/blockchain/android/ui/receive/ReceivePresenter.kt
ditn: Nice, that's better

PR #843, opened 2018-02-12
Now showing correct currency label for Ether

dominicthomas:
- This may not be an issue but was scanning through the file and this caught my eye.. looks like they should be swapped around


ditn:
Have contacted the author in regards to this PR.

PR #848, opened 2018-02-13
Feature/sfox

riaanjvos:


app/src/test/java/piuk/blockchain/android/ui/dashboard/DashboardPresenterTest.kt
```diff
@@ -74,6 +84,29 @@ class DashboardPresenterTest {
                 .thenReturn(Observable.just(emptyList()))
     }
 
+    /*
+    Avoid RuntimeException when using 'observeOn(AndroidSchedulers.mainThread()'
+     */
+    fun setUpRxSchedulers() {
```
ditn: This shouldn't be necessary... I wonder what's happened here 🤔

ditn: If you extend `RxTest` it should be fine I think

riaanjvos: Ah yes you're right, extending RxTest fixed it

PR #861, opened 2018-02-16
Fix/bch

ditn:


app/src/main/java/piuk/blockchain/android/ui/swipetoreceive/SwipeToReceivePresenter.kt
```diff
@@ -72,7 +72,7 @@ class SwipeToReceivePresenter @Inject constructor(
             view.setUiState(UiState.EMPTY)
         } else {
             single.doOnSuccess { require(it.isNotEmpty()) { "Returned address is empty, no more addresses available" } }
-                    .doOnSuccess { view.displayReceiveAddress(it) }
+                    .doOnSuccess { view.displayReceiveAddress(it.replace("bitcoincash:", "")) }
```
riaanjvos: This will work, but we should start using FormatsUtil.toShortCashAddress

PR #866, opened 2018-02-19
Fixes/6 10 0 rc1

riaanjvos:


ditn:
Tests aren't building - `WalletOptionsDataManagerTest.kt: (37, 108): No value passed for parameter 'environmentSettings'`

PR #888, opened 2018-02-28
Hotfix/bch

ditn:
Fixes for the most common issues in the 6.10.0 release


app/src/main/java/piuk/blockchain/android/ui/base/BasePresenter.java
```diff
@@ -30,6 +30,7 @@ public void initView(VIEW view) {
 
     @Override
     public VIEW getView() {
+        // TODO: 28/02/2018 In the future, migrate this to Kotlin and mark as nullable
```
riaanjvos: Good idea

PR #896, opened 2018-03-06
Feature/support fonts

ditn:
This pull request adds support for Downloadable Fonts, and removes the old font files from the APK. This allows us to not rely on another 3rd party library (as good as Calligraphy is), but also removes unneeded bloat from the app. Instead, Android itself will download and cache the fonts for us.

5.4MB -> 5.0MB, a 7.5% reduction in download size.


app/src/main/res/font/montserrat_light.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<font-family xmlns:app="http://schemas.android.com/apk/res-auto"
+    app:fontProviderAuthority="com.google.android.gms.fonts"
+    app:fontProviderPackage="com.google.android.gms"
+    app:fontProviderQuery="name=Montserrat&amp;weight=300"
```
CoreFloDev: Hi Adam, I have read your article https://medium.com/@iateyourmic/migrating-to-downloadable-fonts-on-android-bff3fd8d9ae6 which is quite nice thanks for that!

But I don't understand how comes this part work for you, in my case with `name=Montserrat&amp;weight=300` it select `name=Montserrat` instead.

ditn: It's possible that the API has changed or the font provider is different - it's been a long time since I've looked at this! It was definitely working correctly when I implemented it: I remember being surprised that it worked 😅

PR #902, opened 2018-03-19
Refactor/monetary util

riaanjvos:
This is quite a big PR.
Currently most of the presenters handled fetching of exchange rates, doing some multiplication, figuring out which decimal format needs to be used, figure out what currency symbol needs to be used, and then mashes it all together - This worked, but was overkill.
In this PR there is a new Class called CurrencyFormatManager which can be injected into any presenter and will handle all of the above calculations.


app/src/main/java/piuk/blockchain/android/ui/balance/adapter/FctxDelegate.kt
```diff
@@ -202,29 +192,31 @@ class FctxDelegate<in T>(
...
+//            spannable = Spannable.Factory.getInstance().newSpannable(
+//                    "${monetaryUtil.getFiatFormat(fiatString).format(Math.abs(fiatAmount))} $fiatString")
+//            spannable.setSpan(
+//                    RelativeSizeSpan(0.67f),
+//                    spannable.length - 3,
+//                    spannable.length,
+//                    Spannable.SPAN_EXCLUSIVE_EXCLUSIVE)
+//        }
+        spannable = Spannable.Factory.getInstance().newSpannable(
+                "fix me")
```
ditn: Looks like this was missed 😆

ditn: Although I guess it's contacts related, so it's not very important

riaanjvos: Ha you're right. Had a look at it this morning. It's not a quick fix for code we aren't using/potentially needs a rewrite anyway. Okay to skip this?

PR #932, opened 2018-04-18
Feature/buysell welcome routing

riaanjvos:


buysellui/src/main/java/piuk/blockchain/androidbuysellui/injector/BuySellApplicationModule.java
```diff
@@ -0,0 +1,14 @@
...
+
+import dagger.Module;
+import dagger.Provides;
+import info.blockchain.wallet.payload.PayloadManager;
+
+@Module
+public class BuySellApplicationModule {
+
+    @Provides
+    protected PayloadManager providePayloadManager() {
```
ditn: This will work for now but this isn't a sustainable solution haha. Let's look at this tomorrow.

buysellui/src/test/java/piuk/blockchain/androidbuysellui/RxTest.kt
```diff
@@ -0,0 +1,49 @@
...
+import io.reactivex.schedulers.TestScheduler
+import org.junit.After
+import org.junit.Before
+
+/**
+ * Class that forces all Rx observables to be subscribed and observed in the same thread through the
+ * same Scheduler that runs immediately. Also exposes a [TestScheduler] for testing of
+ * time-based methods.
+ */
+open class RxTest {
```
ditn: It's super annoying that we've had to do this in every module. Let's move RxTest and some utils to a separate testing module, which we can share across all other modules.

buysellui/src/main/java/piuk/blockchain/androidbuysellui/injector/BuySellApplicationModule.java
riaanjvos: Yeah I know. Was pulling my hair out over this one. Eventually just put this in so I can continue with the UI work

PR #940, opened 2018-04-24
Feature/coinify apis

ditn:


ditn:
I actually don't think this diff has rendered properly - this PR is quite minor. GitHub might be freaking out a bit

PR #957, opened 2018-05-04
Feature/buysell chrome custom tab

riaanjvos:


app/src/main/java/piuk/blockchain/android/ui/buysell/coinify/signup/identity_in_review/CoinifyIdentityInReviewPresenter.kt
```diff
@@ -0,0 +1,110 @@
...
+            view.onShowDocumentsRequested()
+        } else if (kycList.find{ it.state == ReviewState.Expired } != null) {
+            // Unlikely to see this result - User would be redirected to supply docs again before getting to this fragment
+            view.onShowExpired()
+        } else if (kycList.find{ it.state == ReviewState.Failed } != null) {
+            // We get stuck with the below cases
+            // Can't create new account with same email
+            // redirectUrl isn't valid - Same issue on web
+            view.onShowFailed()
+        } else if (kycList.find{ it.state == ReviewState.Rejected } != null) {
```
ditn: I think you can use `if (kycList.any { it.state == ReviewState.Rejected })` here instead if you want

PR #978, opened 2018-07-04
AND-1186 Rename base @Before method that was being hidden

alanevansbc:
The cause of the flakey tests I think was this base setup method was never called due to being overridden. Sometimes they passed when another test ran earlier that successfully set the blockchain framework.

No doubt we still want to improve this so that we don't have a static init method as part of AND-1184 but this should stop the failures for now.

### DIFF
--- a/wallet/src/test/java/info/blockchain/wallet/MockedResponseTest.java
+++ b/wallet/src/test/java/info/blockchain/wallet/MockedResponseTest.java
@@ -26,7 +26,7 @@ public abstract class MockedResponseTest {
     public MockInterceptor mockInterceptor = MockInterceptor.getInstance();
 
     @Before
-    public void setUp() throws Exception {
+    public void initBlockchainFramework() throws Exception {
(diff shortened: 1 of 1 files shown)


wallet/src/test/java/info/blockchain/wallet/MockedResponseTest.java
```diff
@@ -26,7 +26,7 @@
     public MockInterceptor mockInterceptor = MockInterceptor.getInstance();
 
     @Before
-    public void setUp() throws Exception {
+    public void initBlockchainFramework() throws Exception {
```
alanevansbc: Descendant test classes like `SharedMetadataTest` also have a `setUp` method but don't call `super()`.

ditn: I know there's an Android Support Annotation `@CallSuper`; I don't know if we've got something similar on the classpath in this particular module.

PR #979, opened 2018-07-04
fix(Fabric issue) AND-1187 #1449 Issue: java.lang.IllegalStateException: payloadManager.payload must not be null

robertogdsbc:
AND-1187
Issue: java.lang.IllegalStateException: payloadManager.payload must not be null
Following the approach we used on getHDAccounts() I added checking if payload is null, and then use an empty list. This is to avoid the crash in case that walletBaseBody is null, as we are returning null in PayloadManager->getPayLoad()

### DIFF
--- a/app/src/main/java/piuk/blockchain/android/ui/receive/WalletAccountHelper.kt
+++ b/app/src/main/java/piuk/blockchain/android/ui/receive/WalletAccountHelper.kt
@@ -26,13 +26,13 @@ import javax.inject.Inject
 @Mockable
 @PresenterScope
 class WalletAccountHelper @Inject constructor(
-        private val payloadManager: PayloadManager,
-        private val stringUtils: StringUtils,
-        private val currencyState: CurrencyState,
-        private val ethDataManager: EthDataManager,
-        private val bchDataManager: BchDataManager,
-        private val environmentSettings: EnvironmentConfig,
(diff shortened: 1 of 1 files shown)


app/src/main/java/piuk/blockchain/android/ui/receive/WalletAccountHelper.kt
```diff
@@ -81,27 +81,29 @@ class WalletAccountHelper @Inject constructor(
...
      * Returns a list of [ItemAccount] objects containing only [LegacyAddress] objects.
      *
      * @return Returns a list of [ItemAccount] objects
      */
-    fun getLegacyAddresses() = payloadManager.payload.legacyAddressList
-            // Skip archived address
-            .filterNot { it.tag == LegacyAddress.ARCHIVED_ADDRESS }
+    fun getLegacyAddresses(): List<ItemAccount> {
+        val list = payloadManager.payload?.legacyAddressList
+                ?: Collections.emptyList<LegacyAddress>()
```
ditn: You can just do `emptyList()` here if you so wish but this works fine.

alanevansbc: Yeah would rather see `emptyList()`. But this is not the only instance of `Collections.emptyList` in this file. So it will get cleaned up later when we tighten the ktlint I hope.

PR #983, opened 2018-07-05
AND-1195 fix/improve RX tests

alanevansbc:


app/src/test/java/piuk/blockchain/android/ui/buysell/details/awaitingtransfer/CoinifyAwaitingBankTransferPresenterTest.kt
```diff
@@ -9,33 +9,38 @@ import io.reactivex.Single
...
     private val exchangeService: ExchangeService = mock()
     private val coinifyDataManager: CoinifyDataManager = mock()
     private val view: CoinifyAwaitingBankTransferView = mock()
 
+    @Suppress("unused")
+    @get:Rule
+    val initSchedulers = rxInit {
+        mainTrampoline()
+        ioTrampoline()
+    }
```
alanevansbc: Usage is very targeted and customised for the test in question, rather than the shotgun approach of inheriting from `RxTest`.

Also, here is an example of usage for `TestScheduler`:

```
    val testScheduler = TestScheduler()

    @Suppress("unused")
    @get:Rule
    val initSchedulers = rxInit {
        io(testScheduler)
    }
```

PR #984, opened 2018-07-06
Adam/and 1180

ditn:
This PR enables both `ktlint` and `lint` having set a new baseline.


app/src/main/java/piuk/blockchain/android/ui/buysell/coinify/signup/createaccountstart/CoinifyCreateAccountStartFragment.kt
```diff
@@ -6,7 +6,7 @@ import android.support.v4.app.Fragment
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
-import kotlinx.android.synthetic.main.fragment_coinify_create_account_start.buysellWelcomeButton
+import kotlinx.android.synthetic.main.fragment_coinify_create_account_start.*
```
alanevansbc: How did this get in here? Thought there was a lint rule against `*` imports?

ditn: That's a really good question, let me look into it a second

ditn: It may be a bug in ktlint where a wildcard import passes if there's only one class imported. Seems I still had synthetic imports allowed in AS so that explains why it changed.

ditn: Actually nevermind, it seems it ignores synthetic imports which is probably best given that they're enable by default in AS.

PR #985, opened 2018-07-06
chore(Remove flavour): AND-1197 Remove API level flavour dimension

alanevansbc:


westonal:
Trouble merging, I think because I opened under my other user, so I'm closing and creating new PR

PR #987, opened 2018-07-06
Adam/and 1198

ditn:
This touched quite a few classes but the changes are only minor. Just passing parameters instead of accessing them via Singleton.

### DIFF
--- a/app/src/main/java/piuk/blockchain/android/ui/buy/BuyPresenter.java
+++ b/app/src/main/java/piuk/blockchain/android/ui/buy/BuyPresenter.java
@@ -7,6 +7,7 @@ import javax.inject.Inject;
 import piuk.blockchain.android.R;
 import piuk.blockchain.androidbuysell.datamanagers.BuyDataManager;
 import piuk.blockchain.androidcore.data.access.AccessState;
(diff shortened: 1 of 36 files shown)


app/src/main/java/piuk/blockchain/android/ui/buy/BuyPresenter.java
```diff
@@ -88,7 +92,7 @@ void decryptAndGenerateMetadataNodes(@Nullable String secondPassword) {
             getView().setUiState(UiState.EMPTY);
         } else {
             try {
-                payloadDataManager.decryptHDWallet(secondPassword);
+                payloadDataManager.decryptHDWallet(environmentConfig.getBitcoinNetworkParameters(), secondPassword);
```
alanevansbc: Can the `BuyPresenter` take `NetworkParameters`? or what's the plan when you can buy more than Bitcoin?

ditn: This presenter is for the legacy webview-based buy flow so it doesn't particularly matter. Good spot though!

PR #999, opened 2018-07-11
fix(Currency Selection): AND-1226 Fix for exchange rate not changing …

ditn:
…once loaded

### DIFF
--- a/app/src/main/java/piuk/blockchain/android/ui/buysell/createorder/BuySellBuildOrderPresenter.kt
+++ b/app/src/main/java/piuk/blockchain/android/ui/buysell/createorder/BuySellBuildOrderPresenter.kt
@@ -487,7 +487,9 @@ class BuySellBuildOrderPresenter @Inject constructor(
                         return@BiFunction trader to inMedium
                     }
                 ).flatMap { (trader, inMedium) ->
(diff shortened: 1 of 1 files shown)


app/src/main/java/piuk/blockchain/android/ui/buysell/createorder/BuySellBuildOrderPresenter.kt
```diff
@@ -487,7 +487,9 @@ class BuySellBuildOrderPresenter @Inject constructor(
                         return@BiFunction trader to inMedium
                     }
                 ).flatMap { (trader, inMedium) ->
-                    getExchangeRate(token, -1.0, trader.defaultCurrency)
+                    val currency = if (initialLoad) trader.defaultCurrency else selectedCurrency!!
```
westonal: There are a lot of places in this file where there is `selectedCurrency!!`, how about replacing all of them with a new property:

```
val selectedOrDefaultCurrency: String
   get() = if (initialLoad) trader.defaultCurrency else selectedCurrency!!
```
However, `!!` is a smell, I think we can just do:

```
val selectedOrDefaultCurrency: String
   get() = selectedCurrency ?: trader.defaultCurrency
```

PR #1002, opened 2018-07-12
fix(Currency Selection): AND-1228 Remove USD as a currency option

ditn:


app/src/main/java/piuk/blockchain/android/ui/buysell/createorder/BuySellBuildOrderPresenter.kt
```diff
@@ -78,7 +78,7 @@ class BuySellBuildOrderPresenter @Inject constructor(
             if (isSell) loadMax(new)
         }
     }
-    var selectedCurrency: String by Delegates.observable("USD") { _, old, new ->
+    var selectedCurrency: String by Delegates.observable("EUR") { _, old, new ->
```
westonal: Should/could this use `defaultCurrency`?

Say if you went back to nullable and at places it's used did:

```
selectedCurrency ?: defaultCurrency
```

PR #1004, opened 2018-07-12
refactor(Test): AND-1216 Minimize use of @Mockable to robolectric tests.

westonal:
+ Use mock-maker-inline to allow other mocks.

For AND-1216 I want to close a couple of these classes to make them immutable.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -183,6 +183,10 @@ dependencies {
         exclude group: 'com.google.guava'
         exclude group: 'com.google.protobuf'
     }
+    // For AS running of robolectric
+    testImplementation Libraries.testRunner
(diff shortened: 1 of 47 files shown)


app/build.gradle
```diff
@@ -183,6 +183,10 @@ dependencies {
         exclude group: 'com.google.guava'
         exclude group: 'com.google.protobuf'
     }
+    // For AS running of robolectric
```
westonal: I don't know if you want this comment here.
So I found that the tests wouldn't run in AS: see https://github.com/robolectric/robolectric/issues/3839

ditn: Fine with me.

PR #1007, opened 2018-07-13
fix(Pending KYC): AND-1223 Updated messaging for users who back out o…

ditn:
…f KYC

### DIFF
--- a/app/src/main/java/piuk/blockchain/android/ui/buysell/coinify/signup/identityinreview/CoinifyIdentityInReviewFragment.kt
+++ b/app/src/main/java/piuk/blockchain/android/ui/buysell/coinify/signup/identityinreview/CoinifyIdentityInReviewFragment.kt
@@ -1,7 +1,11 @@
 package piuk.blockchain.android.ui.buysell.coinify.signup.identityinreview
 
 import android.content.Context
+import android.graphics.Paint
 import android.os.Bundle
+import android.support.graphics.drawable.VectorDrawableCompat
+import android.support.v4.graphics.drawable.DrawableCompat
+import android.support.v7.view.ContextThemeWrapper
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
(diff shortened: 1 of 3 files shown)


app/src/main/java/piuk/blockchain/android/ui/buysell/coinify/signup/identityinreview/CoinifyIdentityInReviewFragment.kt
```diff
@@ -90,12 +96,30 @@ class CoinifyIdentityInReviewFragment :
...
+        textviewReviewStatus.apply {
+            text = getString(
+                R.string.buy_sell_review_status,
+                getString(R.string.buy_sell_review_status_in_pending)
+            )
+            val vector = VectorDrawableCompat.create(
+                resources,
+                R.drawable.vector_alert,
+                ContextThemeWrapper(requireActivity(), R.style.AppTheme).theme
+            )!!
```
westonal: `!!` is a smell.

Instead of requesting an NPE here with `!!` you can:

```
            val vector = VectorDrawableCompat.create(
                resources,
                R.drawable.vector_alert,
                ContextThemeWrapper(requireActivity(), R.style.AppTheme).theme
            )?.also {
                DrawableCompat.wrap(it)
                DrawableCompat.setTint(it, getResolvedColor(R.color.primary_navy_medium))
            }
```

And if you really want an exception here, you can throw a descriptive one instead:

```
            val vector = VectorDrawableCompat.create(
                resources,
                R.drawable.vector_alert,
                ContextThemeWrapper(requireActivity(), R.style.AppTheme).theme
            ) ?: throw Exception("Could not create vector drawable")
            DrawableCompat.wrap(vector) //as before
```

ditn: Yeah good point, I'll quickly change this.

PR #1011, opened 2018-07-13
refactor(Balance): AND-1216 Use BalanceReporter adapters and more use…

westonal:
… of CryptoValue over Long/BigInteger/BigDecimal

### DIFF
--- a/app/src/test/java/piuk/blockchain/android/ui/dashboard/DashboardPresenterTest.kt
+++ b/app/src/test/java/piuk/blockchain/android/ui/dashboard/DashboardPresenterTest.kt
@@ -2,20 +2,25 @@ package piuk.blockchain.android.ui.dashboard
 
 import com.nhaarman.mockito_kotlin.any
 import com.nhaarman.mockito_kotlin.anyOrNull
+import com.nhaarman.mockito_kotlin.argThat
 import com.nhaarman.mockito_kotlin.atLeastOnce
 import com.nhaarman.mockito_kotlin.eq
 import com.nhaarman.mockito_kotlin.mock
 import com.nhaarman.mockito_kotlin.verify
 import com.nhaarman.mockito_kotlin.verifyNoMoreInteractions
 import com.nhaarman.mockito_kotlin.whenever
+import info.blockchain.balance.AccountKey
+import info.blockchain.balance.CryptoCurrency
+import info.blockchain.balance.CryptoValue
 import io.reactivex.Completable
 import io.reactivex.Observable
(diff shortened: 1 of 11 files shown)


balance/src/main/kotlin/info/blockchain/balance/CryptoValue.kt
```diff
@@ -1,6 +1,38 @@
...
+
 data class CryptoValue(
     val currency: CryptoCurrency,
-    val amount: Long
-)
+
+    /**
+     * Amount in the smallest unit of the currency, Satoshi/Wei for example.
+     */
+    val amount: BigInteger
```
westonal: `Long` isn't enough for Ether. And we had `BigInteger` in Bitcoin/BitcoinCash places anyway.

testutils/src/main/java/piuk/blockchain/android/testutils/RxTest.kt
```diff
@@ -9,8 +9,9 @@ import org.junit.Rule
  * same Scheduler that runs immediately. Also exposes a [TestScheduler] for testing of
  * time-based methods.
  */
+open class RxTest
 @Deprecated("Implement an rxInit rule locally, specifying just the schedulers the test requires")
-open class RxTest {
+constructor() {
```
westonal: By deprecating the constructor I halved the number of warnings, we were getting one for imports as well I noticed.

app/src/test/java/piuk/blockchain/android/ui/dashboard/DashboardPresenterTest.kt
```diff
@@ -569,8 +566,7 @@ class DashboardPresenterTest : RxTest() {
         verify(ethDataManager).fetchEthAddress()
         verify(payloadDataManager).updateAllBalances()
         verify(payloadDataManager).updateAllTransactions()
-        verify(transactionListDataManager).getBtcBalance(any())
-        verify(transactionListDataManager).getBchBalance(any())
+        verifyBalanceQueriesForBtcAndBch()
```
westonal: I think this test over-verifies. There is usually no need to both setup the mock response and verify it was called but I've kept both for now.
It's so hard to change the presenter due to the tests, test driving a new Presenter class from scratch would be my recommendation next time we need to do significant changes to this class.

ditn: Yeah I agree, the tests for this class have been very brittle for some time. Needs a rethink.

PR #1015, opened 2018-07-16
AND-1216 remove duplicated price code

westonal:
- Replace repeated price and format code with code that uses CryptoCurrency enum. 
- Deprecate some more methods that do not take a CryptoCurrency or CryptoValue.

### DIFF
--- a/app/src/main/java/piuk/blockchain/android/ui/chooser/AccountChooserPresenter.kt
+++ b/app/src/main/java/piuk/blockchain/android/ui/chooser/AccountChooserPresenter.kt
@@ -1,5 +1,7 @@
 package piuk.blockchain.android.ui.chooser
 
+import info.blockchain.balance.CryptoCurrency
+import info.blockchain.balance.CryptoValue
 import info.blockchain.wallet.contacts.data.Contact
 import info.blockchain.wallet.payload.data.LegacyAddress
 import io.reactivex.Observable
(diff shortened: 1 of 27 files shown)


core/src/main/java/piuk/blockchain/androidcore/data/currency/CurrencyFormatManager.kt
```diff
@@ -81,34 +81,10 @@ class CurrencyFormatManager @Inject constructor(
...
+        getFormattedCoinValue(CryptoValue(currencyState.cryptoCurrency, coinValue))
 
-        return when (currencyState.cryptoCurrency) {
-            CryptoCurrency.BTC -> currencyFormatUtil.formatBtc(convertedCoinValue)
-            CryptoCurrency.ETHER -> currencyFormatUtil.formatEth(convertedCoinValue)
-            CryptoCurrency.BCH -> currencyFormatUtil.formatBch(convertedCoinValue)
-            else -> throw IllegalArgumentException(currencyState.cryptoCurrency.toString() + " not supported.")
-        }
-    }
+    fun getFormattedCoinValue(cryptoValue: CryptoValue) = currencyFormatUtil.format(cryptoValue)
```
westonal: Actually, this doesn't format ETH the same way. I will need to write some tests around this as this didn't get flagged up during refactor.

westonal: On the other hand, this never gets called with non-null `convertEthDenomination`. I.e. I don't think it's ever used for ETH.

app/src/main/java/piuk/blockchain/android/ui/chooser/AccountChooserPresenter.kt
```diff
@@ -282,12 +284,9 @@ class AccountChooserPresenter @Inject internal constructor(
...
 
-    private fun getBtcBalanceString(isBtc: Boolean, btcBalance: Long): String {
-        return if (isBtc) {
-            currencyFormatManager.getFormattedBtcValueWithUnit(
-                btcBalance.toBigDecimal(),
-                BTCDenomination.SATOSHI
-            )
+    private fun getBalanceString(displayAsBtc: Boolean, btcBalance: BigInteger): String {
+        return if (displayAsBtc) {
+            currencyFormatManager.getFormattedValueWithUnit(CryptoValue(CryptoCurrency.BTC, btcBalance))
```
westonal: This is a bug, but I think is deadcode, I will fix by removing in separate PR.

PR #1018, opened 2018-07-17
refactor(Balance): AND-1216 Add FiatValue type

westonal:


app/src/main/java/piuk/blockchain/android/ui/buysell/createorder/BuySellBuildOrderPresenter.kt
```diff
@@ -544,8 +545,8 @@ class BuySellBuildOrderPresenter @Inject constructor(
     }
 
     private fun updateSendAmount(quoteAmount: Double) {
-        val formatted = currencyFormatManager
-            .getFiatFormat(selectedCurrency).format(quoteAmount)
+        val formatted = FiatValue(selectedCurrency, quoteAmount.toBigDecimal())
+            .toStringWithSymbol(view.locale)
```
westonal: @ditn I'm going to push new commit. I pretty sure I changed this from no symbol to with symbol. Can you confirm, should have no symbol here?

westonal: No this is correct, `getFiatFormat` added the currency to the format.

PR #1020, opened 2018-07-18
feat(Balance): AND-1216 Exclude BTC Watch-Only balances from totals

westonal:


wallet/src/main/java/info/blockchain/wallet/payload/data/WalletExtensions.kt
```diff
@@ -0,0 +1,17 @@
...
+        .toSet()
+        .toList()
+
+fun Wallet.allSpendableAccountsAndAddresses(): List<String> =
+    (activeXpubs() + spendableLegacyAddressStrings())
+        .toSet()
+        .toList()
+
+private fun Wallet.activeXpubs(): List<String> =
+    hdWallets?.let { it[0].activeXpubs } ?: emptyList()
```
westonal: To allow me to test drive these nice filters in Kotlin, and not add yet more methods to a large java file I created these extensions. It doesn't look as nice where used from Java, but it will be converted at some point.

PR #1023, opened 2018-07-18
feat(Balance): AND-1216 Add watch only individual balances back to the balance map

westonal:


wallet/src/main/java/info/blockchain/wallet/payload/BalanceManager.java
```diff
@@ -77,6 +79,10 @@ public void updateAllBalances(List<String> legacyAddressList, List<String> allAc
 
                 balanceMap.put(address, balance.getFinalBalance());
 
+                if (excludeFromTotals.contains(address)) {
+                    continue;
+                }
```
westonal: This line/method isn't tested. I'd like to extract this method to a new class with tests. I will open a tech debt ticket for now.

PR #1024, opened 2018-07-18
Adam/and 1256/coinify transaction labels

ditn:
There's undeniably some nastiness in here, but I'm looking to completely overhaul transaction handling in the near future.

### DIFF
--- a/core/src/main/java/piuk/blockchain/androidcore/data/transactions/models/TransactionDtos.kt
+++ b/core/src/main/java/piuk/blockchain/androidcore/data/transactions/models/TransactionDtos.kt
@@ -2,6 +2,7 @@ package piuk.blockchain.androidcore.data.transactions.models
 
 import info.blockchain.wallet.ethereum.data.EthTransaction
 import info.blockchain.wallet.multiaddress.TransactionSummary
+import org.spongycastle.asn1.x500.style.RFC4519Style.name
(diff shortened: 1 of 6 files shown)


core/src/main/java/piuk/blockchain/androidcore/data/transactions/models/TransactionDtos.kt
```diff
@@ -24,10 +24,67 @@ abstract class Displayable {
...
+        result = 31 * result + hash.hashCode()
+        result = 31 * result + inputsMap.hashCode()
+        result = 31 * result + outputsMap.hashCode()
+        result = 31 * result + confirmations.hashCode()
+        result = 31 * result + watchOnly.hashCode()
+        result = 31 * result + doubleSpend.hashCode()
+        result = 31 * result + (totalDisplayableCrypto?.hashCode() ?: 0)
+        result = 31 * result + (totalDisplayableFiat?.hashCode() ?: 0)
+        result = 31 * result + (note?.hashCode() ?: 0)
+        return result
```
westonal: No need to write `equals` and `hashcode` yourself. Just use a `data class`.

ditn: I know, but I think because the superclass for the data classes was an abstract class I was finding that equals wasn't working correctly.

PR #1026, opened 2018-07-19
Adam/and 1255/kyc review status

ditn:
This changes the KYC flow so that users that don't complete the flow or are rejected can still access card payments.


app/src/main/java/piuk/blockchain/android/ui/buysell/overview/CoinifyOverviewPresenter.kt
```diff
@@ -164,6 +164,43 @@ class CoinifyOverviewPresenter @Inject constructor(
...
+                    view.showAlertDialog(R.string.buy_sell_overview_pending_kyc_not_found)
+                }
+            )
+    }
+
+    internal fun onRestartKycSelected() {
+        tokenSingle
+            .flatMap { coinifyDataManager.startKycReview(it) }
+            .subscribeOn(Schedulers.io())
+            .observeOn(AndroidSchedulers.mainThread())
```
westonal: Much nicer and works better with the observe on closer to subscribe 👍

PR #1030, opened 2018-07-19
feat(Non-Spendable): AND-1217 Non-Spendable balance, balance calculation refactors

westonal:


wallet/src/main/java/info/blockchain/wallet/payload/CryptoBalanceMap.kt
```diff
@@ -0,0 +1,63 @@
...
+data class CryptoBalanceMap(
+    private val cryptoCurrency: CryptoCurrency,
+    private val xpubs: Set<String>,
+    private val legacy: Set<String>,
+    private val watchOnlyLegacy: Set<String>,
+    private val balances: Map<String, BigInteger>
+) {
+    val totalSpendable = CryptoValue(cryptoCurrency, (xpubs + legacy - watchOnlyLegacy).sum(balances))
+    val totalSpendableLegacy = CryptoValue(cryptoCurrency, (legacy - watchOnlyLegacy).sum(balances))
+    val totalWatchOnly = CryptoValue(cryptoCurrency, watchOnlyLegacy.sum(balances))
```
westonal: Note that this is an immutable data structure, we never need to worry about any totals getting out of step, the only way to update is to generate a new one, which will recalculate the totals. See how the subtract works below.

PR #1033, opened 2018-07-20
fix(Coinify): AND-1273 Fix incorrect metadata which was causing web t…

ditn:
…o open a sell modal


app/src/main/java/piuk/blockchain/android/ui/buysell/confirmation/buy/CoinifyBuyConfirmationPresenter.kt
```diff
@@ -159,7 +159,7 @@ class CoinifyBuyConfirmationPresenter @Inject constructor(
                     .apply {
                         id = trade.id
                         state = trade.state.toString()
-                        isBuy = false
+                        isBuy = true
```
westonal: Mutation in a map again here. Create ticket to deal with it?

ditn: Yes right you are.

ditn: Ticket made!

PR #1042, opened 2018-07-23
fix(BCH Confirmations): AND-1285 BCH requires 3 confirmations

westonal:
- Consolidated repeated constants.
- Added tests.

### DIFF
--- a/balance/src/main/kotlin/info/blockchain/balance/CryptoCurrency.kt
+++ b/balance/src/main/kotlin/info/blockchain/balance/CryptoCurrency.kt
@@ -3,10 +3,30 @@ package info.blockchain.balance
 import java.math.BigDecimal
 import java.math.BigInteger
 
-enum class CryptoCurrency(val symbol: String, val unit: String, val dp: Int) {
-    BTC("BTC", "Bitcoin", 8),
-    ETHER("ETH", "Ether", 18),
-    BCH("BCH", "Bitcoin Cash", 8);
+enum class CryptoCurrency(
+    val symbol: String,
+    val unit: String,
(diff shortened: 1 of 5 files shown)


balance/src/main/kotlin/info/blockchain/balance/CryptoCurrency.kt
```diff
@@ -3,10 +3,30 @@ package info.blockchain.balance
...
+        unit = "Ether",
+        dp = 18,
+        requiredConfirmations = 12
+    ),
+    BCH(
+        symbol = "BCH",
+        unit = "Bitcoin Cash",
+        dp = 8,
+        requiredConfirmations = 3
+    );
```
westonal: More than 3 arguments gets confusing in code reviews, so started using parameter names.

balance/src/main/kotlin/info/blockchain/balance/CryptoCurrency.kt
```diff
@@ -3,10 +3,30 @@ package info.blockchain.balance
...
+    val symbol: String,
+    val unit: String,
+    val dp: Int,
+    val requiredConfirmations: Int
+) {
+    BTC(
+        symbol = "BTC",
+        unit = "Bitcoin",
+        dp = 8,
+        requiredConfirmations = 3
```
ditn: Neat.

balance/src/main/kotlin/info/blockchain/balance/CryptoCurrency.kt
ditn: Good idea.

PR #1044, opened 2018-07-23
feat(Non-Spendable): AND-1219 English: Relabel Watch only to Non-Spendable

westonal:
image

image

image

image


app/src/main/res/values-cs/strings.xml
```diff
@@ -249,7 +249,7 @@
     <string name="send_to_same_address_warning">Nelze odeslat na stejnou adresu.</string>
     <string name="make_default">Nastavit jako výchozí</string>
     <string name="privx_required">Je vyžadován privátní klíč</string>
-    <string name="watch_only_spend_instructionss">Tato akce vyžaduje privátní klíč pro bitcoinovou adresu %1$s. Naskenujte prosím QR kód.</string>
```
westonal: No translations @ditn 
I fixed this misspelt key and removed another unused.

ditn: Oh yeah duh.

PR #1047, opened 2018-07-24
fix(Sell): AND-1289 Support second password transactions, parse Trans…

ditn:
…ferState of "unknown"


buysell/src/main/java/piuk/blockchain/androidbuysell/models/coinify/CoinifyTrade.kt
```diff
@@ -112,6 +112,10 @@ sealed class TransferState {
 
     // Transfer is in review
     object Reviewing : TransferState()
+
+    // You will potentially see this state shortly after making a sell trade, when the crypto funds have been sent
+    // but Coinify haven't seen it yet
+    object Unknown : TransferState()
```
ditn: Parsed but never actually used. Parse failure was causing the tx list to fail to load shortly after selling BTC. May just remove this class in future.

PR #1054, opened 2018-07-25
AND-1295 move shared code and resources

westonal:
This is work required for moving some buysellui code which is in still in progress but stalled. This work however is harmless and desirable to include ahead of that.

- Moved all vector drawables found in `app` to `coreui`
- Moved some util classes from `app` to `coreui`/`core` Kept package same to minimize changes.
- Deprecated some format util methods


core/src/main/java/piuk/blockchain/android/util/StringUtils.java
```diff
@@ -19,10 +19,16 @@ public String getString(@StringRes int stringId) {
...
 
+    /**
+     * @deprecated Hides warnings/errors about mismatched number of arguments TODO: Inline, AND-1297
+     */
     public String getQuantityString(@PluralsRes int pluralId, int size) {
         return context.getResources().getQuantityString(pluralId, size, size);
     }
 
+    /**
+     * @deprecated Hides warnings/errors about mismatched number of arguments TODO: Inline, AND-1297
```
ditn: Nice, adding the ticket number is here is a good idea.

westonal: Yeah, could be a rule, "No `TODO`s without corresponding ticket reference."

`TODO`'s are great, but they never get done as we only work on tickets. This proves I've also put it in ticket, plus in some cases may help someone locate the area where the work is to be done.

PR #1056, opened 2018-07-25
AND-1241 KYC feature flag

westonal:
- I've gone for a shared feature flag name with homebrew.
- We can make it country specific easily with a replacement implementation of `kyc.FeatureFlag`.
- Same story if we move from `options.json` at some point, it's no problem to create a replacement implementation of `kyc.FeatureFlag`.
- For new classes in the new module, I've started going for a base package of `com.blockchain.kyc`, so no `"piuk"` or `"android"`, bit cleaner and reflecting our domain better. What do you think?


ditn:
Agreed with the domain name, I don't know where `piuk` originated from.

PR #1058, opened 2018-07-26
AND-1249 Shapeshift endpoints injection

westonal:
To remove a barrier to moving from some shapeshift classes, plus injected now in same way as main wallet api.

### DIFF
--- a/app/src/main/java/piuk/blockchain/android/BlockchainApplication.java
+++ b/app/src/main/java/piuk/blockchain/android/BlockchainApplication.java
@@ -67,9 +67,6 @@ public class BlockchainApplication extends Application implements FrameworkInter
     @Inject
     @Named("explorer")
     protected Lazy<Retrofit> retrofitExplorer;
-    @Inject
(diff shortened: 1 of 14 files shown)


app/src/main/java/piuk/blockchain/android/BlockchainApplication.java
```diff
@@ -158,11 +155,6 @@ public Retrofit getRetrofitExplorerInstance() {
         return retrofitExplorer.get();
     }
 
-    @Override
-    public Retrofit getRetrofitShapeShiftInstance() {
-        return retrofitShapeShift.get();
-    }
-
```
westonal: Eventually we can do the same for the other injected `Retrofit`s here as they can drop off the `BlockchainFramework`/`FrameworkInterface`.

ditn: Great, yeah it doesn't make much sense not to provide them this way.

PR #1071, opened 2018-08-01
feat(ShapeShift): AND-1249 Generic Exchange rate MVI

westonal:


balance/src/main/kotlin/info/blockchain/balance/ExchangeRate.kt
```diff
@@ -0,0 +1,73 @@
...
+            }
+        }
+
+        fun inverse(): CryptoToFiat {
+            return CryptoToFiat(to, from, rateInverse)
+        }
+    }
+}
+
+operator fun CryptoValue.times(rate: ExchangeRate.CryptoToCrypto?) =
```
ditn: Neat.

PR #1072, opened 2018-08-01
refactor(Balances): AND-1249 Crypto Currency formatting no longer requiring injection

westonal:


balance/src/main/kotlin/info/blockchain/balance/CryptoCurrencyFormatter.kt
```diff
@@ -0,0 +1,78 @@
...
+
+private fun Double.toPositiveDouble() = Math.max(this, 0.0)
+
+/**
+ * Replace 0.0 with 0 to match web
+ */
+private fun String.toWebZero() = if (this == "0.0" || this == "0.00") "0" else this
+
+private fun createCryptoDecimalFormat(maxDigits: Int) =
+    (NumberFormat.getInstance(Locale.getDefault()) as DecimalFormat).apply {
```
westonal: The one small snag with this is, when changing default locale. The previous implementation had chance to inject a new instance with new decimal formats. If it's an issue, we can use a `map<Locale,DecimalFormat>`.

ditn: QA will definitely catch this at somepoint, but it's a minor issue as far as I'm concerned.

PR #1074, opened 2018-08-02
feat(Homebrew): AND-1239 Add navigation controller + host activity + first fragment of signup flow

ditn:


kyc/src/main/java/com/blockchain/kycui/KycSplashFragment.kt
```diff
@@ -0,0 +1,96 @@
...
+            override fun onClick(view: View) {
+                startActivity(Intent(Intent.ACTION_VIEW, Uri.parse(URL_PRIVACY_POLICY)))
+            }
+        }
+
+        formatLinks(
+            textViewTerms,
+            listOf(disclaimerStart, terms, ampersand, privacy),
+            listOf(defaultClickSpan, termsClickSpan, defaultClickSpan, privacyClickSpan)
+        )
```
westonal: Rather than two distinct lists that contain items associated by their index in the list, you can pair them off:
```
textViewTerms.formatLinks(
    disclaimerStart to defaultClickSpan,
    terms to termsClickSpan,
    ampersand to defaultClickSpan,
    privacy to privacyClickSpan
)
```
Then there is no need for a `zip` or `require` test for different lengths.
```
private fun TextView.formatLinks(
        vararg links: Pair<String, ClickableSpan>
    ) {
...
    links.forEach { (link, span) ->
```

kyc/src/main/java/com/blockchain/kycui/KycSplashFragment.kt
```diff
@@ -0,0 +1,96 @@
...
+        val termsClickSpan = object : ClickableSpan() {
+            override fun onClick(view: View) {
+                startActivity(Intent(Intent.ACTION_VIEW, Uri.parse(URL_TOS_POLICY)))
+            }
+        }
+        val privacyClickSpan = object : ClickableSpan() {
+            override fun onClick(view: View) {
+                startActivity(Intent(Intent.ACTION_VIEW, Uri.parse(URL_PRIVACY_POLICY)))
+            }
+        }
```
westonal: This is repeated, how about a function.

```
 val privacyClickSpan = hyperlinkSpan(URL_PRIVACY_POLICY)
```

kyc/src/main/java/com/blockchain/kycui/KycSplashFragment.kt
ditn: That's a good shout, I'll change this.

PR #1084, opened 2018-08-03
refactor(ShapeShift): AND-1249 Move Account Chooser

westonal:
- I've detangled it from the `ItemAccount` type that is used in many places
- It also now gets it's account list from a simplified interface `AccountListing`, the old implementation is used behind an adapter pattern, some tests are needed here, will add before merge
- I still need to move the activity, it's complex because of injection
- I've had to duplicate some layouts because of koltinx issue


app/src/main/java/piuk/blockchain/android/ui/chooser/WalletAccountHelperAccountListingAdapter.kt
```diff
@@ -0,0 +1,47 @@
...
+
+import info.blockchain.balance.CryptoCurrency
+import io.reactivex.Observable
+import piuk.blockchain.android.ui.account.ItemAccount
+import piuk.blockchain.android.ui.receive.WalletAccountHelper
+import piuk.blockchain.androidcore.injection.PresenterScope
+import javax.inject.Inject
+
+@PresenterScope
+class WalletAccountHelperAccountListingAdapter @Inject constructor(
```
westonal: Gives us new `AccountListing` functionality from old `WalletAccountHelper`.

westonal:
>  ...some tests are needed here, will add before merge

I added them.

PR #1086, opened 2018-08-06
feat(Homebrew): AND-1319 Add Country Selection screen via Fragment

ditn:
screenshot_1533562518


ditn:
Apologies for the number of files touched; there's a few changes here that'll allow me to just crack on getting the UI flow sorted today and tomorrow, and a few files moved.

core/src/main/java/piuk/blockchain/androidcore/utils/Country.kt
```diff
@@ -0,0 +1,31 @@
+package piuk.blockchain.androidcore.utils
+
+import java.util.Locale
+
+class CountryHelper(val locale: Locale) {
```
westonal: I think that this name isn't very OO, as `Helper`s don't generally model anything ([most things ending in `er` are not OO](https://www.yegor256.com/2015/03/09/objects-end-with-er.html)).

Also you don't really need/use this class name, the only usage is:

```
private val countryList = CountryHelper(Locale.getDefault()).countryList
```

Wouldn't it be nice if `.countryList()` was an extension function of `Locale`?

At place of usage, this would be:

```
private val countryList = Locale.getDefault().countryList()
```

ditn: That's a really nice idea. I'll change this a second.

PR #1091, opened 2018-08-08
feat(HomeBrew): AND-1327 Split and refactor EnvironmentConfig and usages

westonal:


app/src/main/java/piuk/blockchain/android/injection/ApiModule.java
```diff
@@ -32,7 +31,7 @@
 import piuk.blockchain.androidbuysell.models.coinify.TransferStateAdapter;
 import piuk.blockchain.androidcore.BuildConfig;
 import piuk.blockchain.androidcore.data.api.ConnectionApi;
-import piuk.blockchain.androidcore.data.api.EnvironmentConfig;
+import piuk.blockchain.androidcore.data.api.EnvironmentUrls;
```
westonal: `EnvironmentConfig` is too much to move, all this module needs is some URLs. Once I have moved this, I can add Koin and the module for this elsewhere and it can be used in all the demo/dev apps too, unlike now.

app/src/main/java/piuk/blockchain/android/data/api/EnvironmentSettings.kt
```diff
@@ -13,6 +14,12 @@ import piuk.blockchain.androidcore.data.api.EnvironmentConfig
 
 class EnvironmentSettings : EnvironmentConfig {
 
+    override fun websocketUrl(currency: CryptoCurrency) = when (currency) {
+        CryptoCurrency.BTC -> BuildConfig.BITCOIN_WEBSOCKET_URL
+        CryptoCurrency.ETHER -> BuildConfig.ETHEREUM_WEBSOCKET_URL
+        CryptoCurrency.BCH -> BuildConfig.BITCOIN_CASH_WEBSOCKET_URL
+    }
```
westonal: Separate PR mainly because of these changes, I don't want them to be hidden.

ditn: Yeah cool, that's sensible.

PR #1093, opened 2018-08-09
feat(HomeBrew): AND-1327 App uses Koin, Dagger uses Koin indirectly

westonal:


app/src/main/java/com/blockchain/koin/KoinStarter.kt
```diff
@@ -0,0 +1,46 @@
...
+import org.koin.standalone.StandAloneContext
+import timber.log.Timber
+
+object KoinStarter {
+
+    private lateinit var application: Application
+
+    @JvmStatic
+    fun start(application: Application) {
+        StandAloneContext.closeKoin()
```
westonal: As an alternative to `extends AutoCloseKoinTest` in every robolectric test.

app/src/test/java/com/blockchain/koin/KoinGraphTest.kt
```diff
@@ -0,0 +1,20 @@
...
+import org.koin.test.AutoCloseKoinTest
+import org.koin.test.dryRun
+import org.robolectric.RobolectricTestRunner
+import org.robolectric.annotation.Config
+import piuk.blockchain.android.BlockchainTestApplication
+import piuk.blockchain.android.BuildConfig
+
+@Config(sdk = [23], constants = BuildConfig::class, application = BlockchainTestApplication::class)
+@RunWith(RobolectricTestRunner::class)
+class KoinGraphTest : AutoCloseKoinTest() {
```
westonal: If graph is wrong, it's possible may robolectric tests will fail too.
Easy way to test just graph without robolectric tests crashing :
```
./gradlew testEnvProdDebug --tests *KoinGraphTest
```

PR #1095, opened 2018-08-09
chore(githooks): AND-000: Set file permissions after installing git hooks

donnfelker:
Marked as `AND-000` as I have no idea what your internal JIRA looks like. 😄 

This simple PR updates your `installGitHooks` command to include setting the file permissions. Therefore, once the command is run, the correct file permissions (executable) will be set on the hooks - preventing manual intervention after the command is run.

### DIFF
--- a/gradle/githooks.gradle
+++ b/gradle/githooks.gradle
@@ -6,6 +6,11 @@ task deleteGitHooksSamples(
     delete '.git/hooks'
 }
 
+task setFilePermissions(type: Exec) {
+    // Mark all files in the .git/hooks directory as executable
+    commandLine 'sh', '-c', 'find .git/hooks -type f -exec chmod 755 {} \\;'
+}
+
 task installGitHooks(
         type: Copy,
         dependsOn: 'deleteGitHooksSamples',
@@ -14,4 +19,7 @@ task installGitHooks(
 ) {
     from 'githooks/'
     into '.git/hooks'
+    doLast {
+      setFilePermissions.execute()
+    }
 }


westonal:
I can't reproduce in `develop` or `master`.
Git stores the execute bit.
So the file in the repo should already be executable.
Please provide reproduction steps.

```
$ git clone git@github.com:blockchain/My-Wallet-V3-Android.git TEST
$ cd TEST
//setup secrets files etc as required
$ ./gradlew installGitHooks
$ git commit -m"sss" --amend
Aborting commit. Commit message does not adhere to standard format:
(feat|fix|docs|style|refactor|test|chore)\((.{3,})\): (AND-\d{3,4}) (.{10,})
```

westonal:
But thanks for the comment on the not being able to see our internal tracking software.

We didn't expect external contributors to worry about the commit message style, as we can always adjust the message before we squash and merge.

But what we have decided, is that we will change the hook to accept both AND- and # github style.

donnfelker:
I see what happened. I cloned the repo in an incorrect drive location and then copied the files to a different directory. Upon copy/pasting the perms were lost. I still feel this change, while not 100% necessary, is useful as anyone who runs `./gradlew installGitHooks` will get the proper hooks installed with correct permissions. Feel free to take it or leave it though! :)

westonal:
Does that count as a "copy and paste" error? 😁 

- we're unsure about making a whole directory executable
- we're not sure what will happen on a windows machine
- we've never had the issue ourselves
- external contributors do not need to worry about the commit message content, as we can clean up at time of "squash and merge"
- we don't merge PRs to master, but develop (that detail should be in our contrib. guidelines)

So for those reasons, I think we will leave it, thanks though.

PR #1096, opened 2018-08-09
 feat(HomeBrew): AND-1327 Inject to Account Chooser with Koin

westonal:
- In order to inject with Koin, all of the dependencies and their dependencies needed to appear in Koin modules.
- I repeatedly ran the KoinGraph test until I had them all.
- What I'm starting to see is that each library can easily have a Koin module that include its local classes.
- So now we have `walletModule`, `coreModule`, and `coreUIModule` that are defined in those libraries, making it easier to reuse in demo/dev apps.
- One gotcha with converting is that `@Singleton @Inject constructor` must be replaced with a `@Provides` calling Koin else you end up with a different singleton on the Koin side to Dagger.


core/src/main/java/piuk/blockchain/androidcore/data/bitcoincash/BchDataStore.kt
```diff
@@ -3,14 +3,11 @@ package piuk.blockchain.androidcore.data.bitcoincash
...
 import info.blockchain.wallet.coin.GenericMetadataWallet
 import piuk.blockchain.androidcore.data.datastores.SimpleDataStore
-import javax.inject.Inject
-import javax.inject.Singleton
 
 /**
  * A simple data store class to cache the Bitcoin cash Wallet (bitcoin chain M/44H/0H)
  */
-@Singleton
-class BchDataStore @Inject constructor() : SimpleDataStore {
```
westonal: These must be carefully converted to `@Provides` delegating to a `bean` as I mentioned in the main body of this PR.

coreui/src/main/java/piuk/blockchain/androidcoreui/injector/ContextModule.kt
```diff
@@ -1,14 +1,63 @@
...
+class ContextModule(private val appContext: Context) : KoinDaggerModule() {
 
     @Singleton
     @Provides
     fun appContext(): Context = appContext
+
+    @Provides
+    fun provideRxBus(): RxBus {
+        return get(RxBus::class)
+    }
```
westonal: These are where I've put the old `@Singleton`s. They don't need to declare as `@Singleton` because we want Koin to do that job (where they are declared as `bean`s) and have a single source of truth.

app/src/main/java/piuk/blockchain/android/ui/chooser/AccountChooserActivity.kt
```diff
@@ -18,14 +19,11 @@ import piuk.blockchain.androidcore.utils.helperfunctions.unsafeLazy
...
 import piuk.blockchain.androidcoreui.utils.extensions.visible
-import javax.inject.Inject
 
 class AccountChooserActivity : BaseMvpActivity<AccountChooserView, AccountChooserPresenter>(),
     AccountChooserView {
 
-    @Suppress("MemberVisibilityCanBePrivate")
-    @Inject
-    lateinit var accountChooserPresenter: AccountChooserPresenter
+    private val accountChooserPresenter: AccountChooserPresenter by inject()
```
westonal: Note this is where I switch from Dagger to Koin for the `AccountChooserActivity`.

PR #1104, opened 2018-08-14
 feat(Homebrew): AND-1235 Create user, fetch user, tests for DataManager

ditn:


core/src/main/java/piuk/blockchain/androidcore/data/api/interceptors/ApiInterceptor.kt
```diff
@@ -23,7 +23,9 @@ class ApiInterceptor : Interceptor {
             request.headers()
         )
 
-        if (request.method().compareTo("post", ignoreCase = true) == 0) {
+        if (request.method().compareTo("post", ignoreCase = true) == 0 ||
+            request.method().compareTo("put", ignoreCase = true) == 0
```
westonal:
```
if (request.method().equals("post", ignoreCase = true) || ...
```

kyc/src/test/java/com/blockchain/kycui/profile/KycProfilePresenterTest.kt
```diff
@@ -78,14 +101,89 @@ class KycProfilePresenterTest {
...
+        verify(view).dismissProgressDialog()
+        verify(view).showErrorToast(any())
+    }
+
+    @Test
+    fun `on continue clicked all data correct, metadata fetch success`() {
+        // Arrange
+        val firstName = "Adam"
+        val lastName = "Bennett"
+        val dateOfBirth = Calendar.getInstance()
```
westonal: I wouldn't use current date in a test.
Timezones and stuff like that can make tests like this flakey.
Instead specify a date here, and use the expected string, not `dateOfBirth.toSimpleDateString()`.

core/src/main/java/piuk/blockchain/androidcore/data/api/interceptors/ApiInterceptor.kt
ditn: Oh yes obviously, I'm not sure how I didn't spot that when this class was converted from Java.

kyc/src/test/java/com/blockchain/kycui/profile/KycProfilePresenterTest.kt
ditn: I figured I could get away with it because it's not being compared to anything, but you're probably right. I'll change this quickly.

westonal:
```
whenever(
            nabuDataManager.createBasicUser(
                firstName,
                lastName,
                dateOfBirth.toSimpleDateString(),
                offlineToken.mapFromMetadata()
            )
```
The comparison is here, if it doesn't match up, the `whenever` won't work.

PR #1106, opened 2018-08-14
feat(Homebrew): AND-1333 Refactor countries list to load countries fr…

ditn:
…om endpoint, then check selected country against cached list

### DIFF
--- a/kyc/src/main/java/com/blockchain/kyc/datamanagers/nabu/NabuDataManager.kt
+++ b/kyc/src/main/java/com/blockchain/kyc/datamanagers/nabu/NabuDataManager.kt
@@ -6,6 +6,7 @@ import com.blockchain.kyc.models.nabu.NabuErrorCodes
 import com.blockchain.kyc.models.nabu.NabuOfflineTokenResponse
 import com.blockchain.kyc.models.nabu.NabuSessionTokenResponse
 import com.blockchain.kyc.models.nabu.NabuUser
+import com.blockchain.kyc.models.nabu.Scope
 import com.blockchain.kyc.services.nabu.NabuService
 import com.blockchain.kyc.stores.NabuSessionTokenStore
 import io.reactivex.Completable
@@ -102,12 +103,8 @@ class NabuDataManager(
         nabuTokenStore.invalidate()
     }
 
-    internal fun isInEeaCountry(countryCode: String): Single<Boolean> =
-        nabuService.getEeaCountries()
-            .map { it.containsCountry(countryCode) }
-
-    private fun List<NabuCountryResponse>.containsCountry(countryCode: String): Boolean =
-        this.any { it.code.equals(countryCode, ignoreCase = true) }
+    internal fun getCountriesList(scope: Scope): Single<List<NabuCountryResponse>> =
+        nabuService.getCountriesList(scope = scope)
 
     private fun unauthenticated(it: Throwable) =
         (it as? NabuApiException?)?.getErrorCode() == NabuErrorCodes.TokenExpired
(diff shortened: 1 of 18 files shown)


kyc/src/main/java/com/blockchain/kycui/countryselection/KycCountrySelectionFragment.kt
```diff
@@ -47,44 +49,61 @@ class KycCountrySelectionFragment :
...
-                        it.name.contains(query, ignoreCase = true) ||
-                            it.countryCode.contains(query, ignoreCase = true)
-                    }
+                countryList?.let { list ->
+                    countryCodeAdapter.items =
+                        list.filter {
+                            it.name.contains(query, ignoreCase = true) ||
+                                it.countryCode.contains(query, ignoreCase = true)
+                        }
+                }
```
westonal: You have an empty subscribe and these `doOnNext`, this can all happen in subscribe.

But also, the way you check for null (`countryList?`) suggests we could be subject to race conditions here.

Every time you have a side effect to set some mutable state, you risk race conditions and need complicated code to avoid it.

I think we should be able to achieve something like:

```
disposable += Observable.combineLatest(countryList, userInput)
                         .subscribe { (list, query) -> countryCodeAdapter.items = /* as above */ }
```

In other words: "I don't care what comes in first, when they are both ready, and every time one changes, give me the pair of them and I'll update the adapter."

Then, none of this logic for pairing and searching needs to be in the `KycCountrySelectionFragment`. It's too close to the UI.

It can all be test driven into a class used from fragment like so:

```
disposable += TheExternalClass(countryList, userInput) // inputs still observables
                         .matchingItems //search logic here, very testable
                         .observeOn(AndroidSchedulers.mainThread())
                         .subscribe { countryCodeAdapter.items = it }
```

kyc/src/main/java/com/blockchain/kycui/search/ListQueryObservable.kt
```diff
@@ -0,0 +1,19 @@
...
+
+import io.reactivex.Observable
+import io.reactivex.functions.BiFunction
+
+class ListQueryObservable<T>(
+    private val queryObservable: Observable<CharSequence>,
+    private val listObservable: Observable<List<T>>
+) {
+
+    fun <T> matchingItems(
```
westonal: This `<T>` is giving you that issue where you need to unchecked cast `as List<T>`. Just remove it, this will then use the `T` from the class.

ditn:
Will retarget on top of #1109 once it's merged.

kyc/src/main/java/com/blockchain/kyc/datamanagers/nabu/NabuDataManager.kt
```diff
@@ -72,26 +73,19 @@ class NabuDataManager(
         offlineTokenResponse: NabuOfflineTokenResponse
     ): Completable =
         authenticate(offlineTokenResponse) { session ->
-            emailSingle.flatMapCompletable { email ->
```
ditn: Email is no longer required/allowed for this PUT request.

kyc/src/main/java/com/blockchain/kyc/services/nabu/NabuService.kt
```diff
@@ -75,33 +76,31 @@ class NabuService(
...
-        path: String = apiPath + NABU_USERS,
-        userId: String,
+        path: String = apiPath + NABU_USERS_CURRENT,
         firstName: String,
         lastName: String,
-        email: String,
         dateOfBirth: String,
         sessionToken: String
     ): Completable = service.createBasicUser(
-        "$path/$userId",
```
ditn: `userId` is encoded in the session token, so no longer needed here either.

PR #1107, opened 2018-08-15
AND-1301 inject payload manager

westonal:


wallet/src/main/java/com/blockchain/koin/walletModule.kt
```diff
@@ -1,9 +1,32 @@
...
+
+        bean { PayloadManager(get(), get(), get(), get()) }
+    }
+
+    bean {
+        object : PayloadManagerWiper {
+            override fun wipe() {
+                (StandAloneContext.koinContext as KoinContext).releaseContext("Payload")
+            }
+        } as PayloadManagerWiper
```
westonal: This "wiper" replaces the `wipe` method on `PayloadManager`, which was a `static` mutable.

Now `PayloadManager` has more immutable fields, and no `wipe` method. We just release the "Payload" context and the `PayloadManager` **and everything that depends on it** is gone from the DI containers.

Note that if something is declared outside of the `"Payload"` context, and it depends on `PayloadManager` or anything else inside the context, then `dryRun` fails, and so our tests fail.

westonal: So it is far safer than before, where it relied on developers to clear the relevant fields inside the wipe methods.

wallet/src/test/java/com/blockchain/koin/modules/PayloadManagerWipingTest.kt
```diff
@@ -0,0 +1,42 @@
...
+
+        firstPayloadManager `should be` secondPayloadManager
+
+        val thirdPayloadManager: PayloadManager by inject()
+
+        val wiper: PayloadManagerWiper by inject()
+
+        wiper.wipe()
+
+        thirdPayloadManager `should not be` secondPayloadManager
```
westonal: And I tested the `wipe` behaves as expected here. After wipe, we get a new instance.

wallet/src/test/java/info/blockchain/wallet/payload/PayloadManagerTest.java
```diff
@@ -37,12 +39,17 @@
...
+                BlockchainFramework.getRetrofitExplorerInstance(),
+                BlockchainFramework.getRetrofitApiInstance(),
+                BlockchainFramework.getApiCode()
+        );
+        payloadManager = new PayloadManager(
+                walletApi,
+                new MultiAddressFactory(blockExplorer),
+                new BalanceManagerBtc(blockExplorer),
+                new BalanceManagerBch(blockExplorer)
+        );
```
westonal: Ideally these will come from a Koin module, even a local one. But would need to convert to Kt first.

PR #1108, opened 2018-08-15
refactor(Tests): AND-1258 No shared data between mock interceptors

westonal:
- fresh mock interceptor for each test
- inject FeeApi dependencies


westonal:
Will retarget once #1107 is in

PR #1109, opened 2018-08-15
feat(Homebrew): AND-1333 Test driven country filter

westonal:
This is what I was suggesting, test drive the whole filtering logic, rather than untested logic in a `Fragment`.


kyc/src/main/java/com/blockchain/kycui/countryselection/KycCountrySelectionFragment.kt
```diff
@@ -60,13 +60,7 @@ class KycCountrySelectionFragment :
...
             .debounce(100, TimeUnit.MILLISECONDS)
 
-        compositeDisposable += ListQueryObservable(userInput, countryList)
-            .matchingItems<CountryDisplayModel> { query, list ->
-                list.filter {
-                    it.name.contains(query, ignoreCase = true) ||
-                        it.countryCode.contains(query, ignoreCase = true)
-                }
-            }
+        compositeDisposable += countryList.filterCountries(userInput)
```
westonal: Rather than create a class, I kept it fluent and used extension method here.

PR #1111, opened 2018-08-17
feat(HomeBrew): AND-1336 Quote MVI mappings

westonal:


settings.gradle
```diff
@@ -11,5 +11,6 @@ include ':app',
         ':kycdemoapp',
         ':morph:common',
         ':morph:shapeshift',
+        ':morph:homebrew',
```
ditn: Module all the things!

balance/src/main/kotlin/info/blockchain/balance/CryptoCurrencyFormatter.kt
```diff
@@ -69,10 +81,10 @@ private fun Double.toPositiveDouble() = Math.max(this, 0.0)
 /**
  * Replace 0.0 with 0 to match web
  */
-private fun String.toWebZero() = if (this == "0.0" || this == "0.00") "0" else this
+private fun String.toWebZero() = if (this == "0.0" || this == "0,0" || this == "0.00") "0" else this
```
ditn: Ah good catch.

westonal: Not sure when the `0.00` case kicks in... possibly a old fiat case, if it used to be used for that as well. Anyway without test case, didn't want to add `0,00`. And without knowing what for, didn't want to remove `0.00`! So unbalanced for now :(

PR #1118, opened 2018-08-20
Adam/and 1238/verify phone number

ditn:
screenshot_1534799917


kyc/src/main/java/com/blockchain/kycui/mobile/validation/KycMobileValidationFragment.kt
```diff
@@ -0,0 +1,167 @@
...
+            .skip(1)
+            .observeOn(AndroidSchedulers.mainThread())
+            .map { mapToCompleted(it) }
+            .distinctUntilChanged()
+            .doOnNext {
+                updateProgress(it, kycStep)
+                buttonNext.isEnabled = it
+            }
+
+    private fun mapToCompleted(text: String): Boolean = text.length >= 5
```
westonal: Magic number. I suspect you want to use the `PhoneNumber` class here.

ditn: Ah this is for the validation code. Should encapsulate it though, good shout.

PR #1120, opened 2018-08-21
chore(Gradle): AND-1341 Remove allopen and unnecessary Kapts

westonal:
- `:app` has the only `kapt` usage after this.
- `Mockable` and `allopen` was made defunct with `mock-maker-inline` a while back.


westonal:
I don't think you would notice on app, I did a few experiments with the kyc demo app:

[Kapt times for core and Kyc](https://scans.gradle.com/s/3brkl7t3jptie/performance/execution):

image

Remember these could be in parallel so not saving 7 seconds potentially.

ditn:
Hmm well it's a step closer to removing kapt completely so that's great.

PR #1124, opened 2018-08-21
feat(HomeBrew): AND-1344 Generalize shape shift trade detail activity

westonal:
- Part 1
- Breaking the dependencies of ShapeShift Detail presenter on ShapeShift TDOs, so can use same for HomeBrew

### DIFF
--- a/app/src/main/java/piuk/blockchain/android/injection/ContextModule.kt
+++ b/app/src/main/java/piuk/blockchain/android/injection/ContextModule.kt
@@ -10,6 +10,7 @@ import piuk.blockchain.androidcore.data.contacts.datastore.PendingTransactionLis
 import piuk.blockchain.androidcore.data.ethereum.datastores.EthDataStore
 import piuk.blockchain.androidcore.data.exchangerate.datastore.ExchangeRateDataStore
 import piuk.blockchain.androidcore.data.rxjava.RxBus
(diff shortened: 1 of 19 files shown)


app/src/main/java/piuk/blockchain/android/injection/ContextModule.kt
```diff
@@ -26,6 +27,11 @@ class ContextModule(private val appContext: Context) : KoinDaggerModule() {
         return get(RxBus::class)
     }
 
+    @Provides
+    fun provideShapeShiftDataStore(): ShapeShiftDataStore {
+        return get(ShapeShiftDataStore::class)
```
westonal: A former `@Singleton`, we now want to control life span from Koin.

morph/shapeshift/src/main/java/com/blockchain/koin/modules/shapeshiftModule.kt
```diff
@@ -24,4 +28,13 @@ val shapeShiftModule = applicationContext {
     }
 
     factory { ShapeShiftApi(get()) }
+
+    context("Payload") {
+
+        bean { ShapeShiftDataStore() }
```
westonal: One of these singletons that now benefits from being forced to be in the Payload. It will clear automatically on sign out. But it's safest to keep the `clear` methods around as long as we have dagger.

PR #1129, opened 2018-08-23
feat(HomeBrew): AND-1344 generalize and move state selection activity

westonal:
- Turned out to not be all that necessary a refactor if we won't be doing this, but it was complicating move of the main ShapeShift activity nonetheless.
- I have tested it still works.


morph/ui/src/main/java/com/blockchain/morph/ui/regulation/stateselection/UsStateSelectionActivity.kt
```diff
@@ -8,40 +8,32 @@ import android.view.ViewGroup
...
 import piuk.blockchain.androidcoreui.ui.base.BaseMvpActivity
 import piuk.blockchain.androidcoreui.utils.extensions.gone
 import piuk.blockchain.androidcoreui.utils.extensions.invisible
 import piuk.blockchain.androidcoreui.utils.extensions.visible
-import javax.inject.Inject
 
-class ShapeShiftStateSelectionActivity :
-    BaseMvpActivity<ShapeShiftStateSelectionView, ShapeShiftStateSelectionPresenter>(),
-    ShapeShiftStateSelectionView {
+internal class UsStateSelectionActivity :
```
westonal: Note how activity is `internal` now.

morph/ui/src/main/java/com/blockchain/morph/ui/regulation/stateselection/UsStateSelectionActivity.kt
```diff
@@ -110,15 +104,12 @@ class ShapeShiftStateSelectionActivity :
 
     companion object {
 
-        // TODO we need a request code handler/incrementer to avoid potential collisions
```
westonal: I address this TODO

morph/ui/src/main/java/com/blockchain/koin/morphUiModule.kt
```diff
@@ -0,0 +1,31 @@
...
+
+val morphUiModule = applicationContext {
+
+    bean {
+        object : UsStateSelectionActivityStarter {
+
+            private var requestCode = 55000
+
+            override fun startForResult(parent: Activity): Int {
+                val requestCode = requestCode++
```
westonal: Not sure it was necessary, but here's where I addressed that TODO.

ditn: I'm not sure this is necessary at all - I think Riaan was getting at the fact that some pages use the same request code, but I don't think that's a problem unless in very specific situations.

PR #1132, opened 2018-08-23
refactor(Auth): AND-1335 New nabu module and moved nabu meta data

westonal:
- New module `:nabu`
- Just moving metadata caused me to move all of this serialization code
- I also added tests to remove a TODO


wallet/src/main/java/info/blockchain/wallet/metadata/Saveable.java
```diff
@@ -1,22 +0,0 @@
...
-
-import com.fasterxml.jackson.annotation.JsonIgnore;
-import com.fasterxml.jackson.core.JsonProcessingException;
-
-import java.io.IOException;
-
-public interface Saveable {
-
-    // TODO: 19/04/2018 I'm not sure that this will actually be ignored in a subtype
-    // Requires testing as not to break metadata
```
westonal: This is the TODO

wallet/src/main/java/info/blockchain/wallet/metadata/Saveable.java
```diff
@@ -1,22 +0,0 @@
...
-    @JsonIgnore
-    int getMetadataType();
-
-    @JsonIgnore
-    Saveable fromJson(String json) throws IOException;
```
westonal: This is gone, was not used, and doesn't make sense to have to have an instance before you can deserialize.

common/network/src/test/kotlin/com/blockchain/serialization/SerialisationUtilsTest.kt
```diff
@@ -10,8 +10,8 @@ import org.junit.Test
 class SerialisationUtilsTest {
 
     @Test
-    fun `string to moshi object should return object`() {
-        val testClass = TEST_CLASS_STRING.toMoshiKotlinObject<TestClass>()
+    fun `class fromJsonString should return object`() {
+        val testClass = TestClass::class.fromMoshiJson(TEST_CLASS_STRING)
```
westonal: I have flipped this. Part of this PR makes sure that we don't see serialization when it doesn't apply.
So rather than being able to see `toMoshiKotlinObject` on every `String`. You now need to start with a `KClass` that supports `JsonSerializable`.

common/network/src/main/kotlin/com/blockchain/serialization/Saveable.kt
```diff
@@ -0,0 +1,11 @@
+package com.blockchain.serialization
+
+/**
+ * Exposes a meta data code which is not serialized.
+ */
+interface Saveable : JsonSerializable {
+
+    fun getMetadataType(): Int
```
westonal: The `fun` is required. A `val` gets serialized. The tests cover this to prevent refactoring errors.

common/network/src/test/kotlin/com/blockchain/serialization/SerialisationUtilsTest.kt
ditn: Cool, this is a nice improvement.

wallet/src/main/java/info/blockchain/wallet/metadata/Saveable.java
ditn: Interesting, I thought this was used somewhere but obviously not. I think maybe it's equivalent is used in Jackson Java classes somewhere deep in Metadata in `:wallet`.

PR #1133, opened 2018-08-23
refactor(Auth): AND-1335 Move Nabu Token Store and token models

westonal:


app/src/main/java/piuk/blockchain/android/ui/shapeshift/overview/ShapeshiftPresenter.kt
```diff
@@ -10,7 +10,7 @@ import piuk.blockchain.androidcore.data.currency.CurrencyState
 import piuk.blockchain.androidcore.data.exchangerate.ExchangeRateDataManager
 import piuk.blockchain.androidcore.data.shapeshift.ShapeShiftDataManager
 import piuk.blockchain.androidcore.data.walletoptions.WalletOptionsDataManager
-import piuk.blockchain.androidcore.utils.Optional
+import com.blockchain.utils.Optional
```
ditn: I know we're using this class in a few places but it seems a bit silly. Let's deprecate it in the future and use the Guava (iirc) one instead as it's already on the classpath.

PR #1141, opened 2018-08-29
feat(HomeBrew): AND-1335 auth refactor part3

westonal:
- Add general Nabu authenticator interface - this is not fully generic, but I don't see any value in that just yet and the `Authenticator` interface is in the nabu package.
- After struggling to detach Auth from the data manager (mainly due to tests being hard to refactor), I ended up wrapping it.

### DIFF
--- a/kyc/src/test/java/com/blockchain/kyc/datamanagers/nabu/NabuDataManagerAsAuthenticatorTest.kt
+++ b/kyc/src/test/java/com/blockchain/kyc/datamanagers/nabu/NabuDataManagerAsAuthenticatorTest.kt
@@ -1,0 +1,48 @@
+package com.blockchain.kyc.datamanagers.nabu
+
+import com.blockchain.nabu.Authenticator
+import com.blockchain.nabu.metadata.NabuCredentialsMetadata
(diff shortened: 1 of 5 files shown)


kyc/src/test/java/com/blockchain/kyc/datamanagers/nabu/NabuDataManagerAsAuthenticatorTest.kt
```diff
@@ -0,0 +1,48 @@
...
+        val nabuDataManager = mock<NabuDataManager>()
+        val sut = NabuAuthenticator(metadataManager, nabuDataManager) as Authenticator
+
+        val theFunction = mock<(NabuSessionTokenResponse) -> Single<Int>>()
+        sut.authenticate(theFunction)
+            .test()
+
+        verify(nabuDataManager).authenticate(
+            eq(NabuOfflineTokenResponse("User", "ABC")),
+            any<(NabuSessionTokenResponse) -> Single<Int>>()
```
ditn: I didn't know it was possible to pass functions as a type here but that obviously makes sense. Good to know.

PR #1147, opened 2018-09-03
feat(Homebrew): AND-1380 Sync with Nabu on mobile update

ditn:


kyc/src/main/java/com/blockchain/kyc/models/nabu/NabuErrorCodes.kt
```diff
@@ -21,4 +21,30 @@ enum class NabuErrorCodes(val code: Int) {
...
+    InvalidCredential("INVALID_CREDENTIALS"),
+    WrongPassword("WRONG_PASSWORD"),
+    Wrong2FA("WRONG_2FA"),
+    Bad2FA("BAD_2FA"),
+    UnknownUser("UNKNOWN_USER"),
+    InvalidRole("INVALID_ROLE"),
+    AlreadyLoggedIn("ALREADY_LOGGED_IN"),
+    InvalidStatus("INVALID_STATUS"),
+    BadParamValus("BAD_PARAM_VALUE"),
+    Unknown("UNKNOWN");
```
westonal: What are we doing with all these?

ditn: Currently nothing. But I want to move from the old error codes which are HTTP code based to the new statuses at some point. I thought adding this was necessary for this PR but it turned out not to be, but I left them in anyway.

westonal: image
OK

PR #1148, opened 2018-09-03
feat(HomeBrew): AND-1374 Use existing account picker in HomeBrew

westonal:
- Just logs the result of the selection
- Identical to shapeshift's one, I'd already made this general (available in `:coreui`) restyle can come later, and we should discuss if we even want different user experience from different places in the app. This one is used for sending and receiving as well for example.

image

image

### DIFF
--- a/app/src/main/java/piuk/blockchain/android/ui/account/ItemAccount.kt
+++ b/app/src/main/java/piuk/blockchain/android/ui/account/ItemAccount.kt
@@ -1,5 +1,7 @@
 package piuk.blockchain.android.ui.account
 
+import com.blockchain.serialization.JsonSerializableAccount
+
 @Suppress("LeakingThis")
 class ItemAccount {
 
@@ -12,9 +14,8 @@ class ItemAccount {
     var tag: String? = null
     var absoluteBalance: Long? = null
 
-    // TODO Get rid of this Any
     // Ultimately this is used to sign txs
-    var accountObject: Any? = null
+    var accountObject: JsonSerializableAccount? = null
 
     // Address/Xpub to fetch balance/tx list
     var address: String? = null
@@ -40,7 +41,7 @@ class ItemAccount {
         displayBalance: String?,
         tag: String?,
         absoluteBalance: Long?,
-        accountObject: Any? = null,
+        accountObject: JsonSerializableAccount? = null,
         address: String?,
         type: TYPE = TYPE.SINGLE_ACCOUNT
     ) {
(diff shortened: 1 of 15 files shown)


app/src/main/java/piuk/blockchain/android/ui/account/ItemAccount.kt
```diff
@@ -12,9 +14,8 @@ class ItemAccount {
     var tag: String? = null
     var absoluteBalance: Long? = null
 
-    // TODO Get rid of this Any
```
westonal: I got rid of it!

app/src/main/java/piuk/blockchain/android/ui/shapeshift/newexchange/NewExchangeActivity.kt
```diff
@@ -166,20 +165,14 @@ class NewExchangeActivity : BaseMvpActivity<NewExchangeView, NewExchangePresente
     override fun onActivityResult(requestCode: Int, resultCode: Int, data: Intent?) {
         super.onActivityResult(requestCode, resultCode, data)
         if (resultCode == Activity.RESULT_OK && data != null) {
-
-            val clazz =
-                Class.forName(data.getStringExtra(AccountChooserActivity.EXTRA_SELECTED_OBJECT_TYPE))
-            val any = ObjectMapper().readValue(
-                data.getStringExtra(AccountChooserActivity.EXTRA_SELECTED_ITEM),
-                clazz
-            )
```
westonal: This is repeated in a lot of places, I didn't want to add another.

coreui/src/main/java/com/blockchain/ui/chooser/AccountChooserActivity.kt
```diff
@@ -137,5 +153,30 @@ class AccountChooserActivity : BaseMvpActivity<AccountChooserView, AccountChoose
...
+            return any as? JsonSerializableAccount
+        }
+
+        fun getSelectedAccount(data: Intent): AccountChooserResult {
+            val account = getSelectedRawAccount(data)
+            return when (account) {
+                is Account -> AccountChooserResult(CryptoCurrency.BTC, account)
+                is EthereumAccount -> AccountChooserResult(CryptoCurrency.ETHER, account)
+                is GenericMetadataAccount -> AccountChooserResult(CryptoCurrency.BCH, account)
+                else -> throw IllegalArgumentException("Unsupported class type")
```
westonal: This can simplify the code that uses this selector. They no longer need to do this themselves to discover the currency of an account. This central place will make it easier to add additional currencies.

wallet/src/main/java/info/blockchain/wallet/contacts/data/Contact.java
```diff
@@ -24,7 +25,7 @@
     setterVisibility = Visibility.NONE,
     creatorVisibility = Visibility.NONE,
     isGetterVisibility = Visibility.NONE)
-public class Contact {
+public class Contact implements JsonSerializableAccount {
```
westonal: I suppose "Account" doesn't work here, but Contacts are old and unused, so I'm not too worried.

ditn: Yeah cool. I might talk to Matt about the roadmap and see if we're clear to remove everything.

PR #1151, opened 2018-09-04
feat(Homebrew): AND-1402 Fix back stack handling in KYC flow

ditn:
Not very exciting, but should pop all pages when flow completed and redirect to the status page.

### DIFF
--- a/coreui/src/main/java/com/blockchain/ui/extensions/ThrottledClicks.kt
+++ b/coreui/src/main/java/com/blockchain/ui/extensions/ThrottledClicks.kt
@@ -1,0 +1,11 @@
+package com.blockchain.ui.extensions
+
+import android.view.View
+import com.jakewharton.rxbinding2.view.clicks
+import io.reactivex.Observable
+import io.reactivex.android.schedulers.AndroidSchedulers
+import java.util.concurrent.TimeUnit
+
+fun View.throttledClicks(): Observable<Unit> =
+    clicks()
+        .throttleFirst(500, TimeUnit.MILLISECONDS, AndroidSchedulers.mainThread())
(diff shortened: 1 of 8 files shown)


kyc/src/main/java/com/blockchain/kycui/complete/ApplicationCompleteFragment.kt
```diff
@@ -25,12 +33,29 @@ class ApplicationCompleteFragment : Fragment() {
...
         progressListener.incrementProgress(KycStep.CompletePage)
     }
+
+    override fun onResume() {
+        super.onResume()
+
+        compositeDisposable +=
+            buttonNext
+                .clicks()
+                .debounce(500, TimeUnit.MILLISECONDS, AndroidSchedulers.mainThread())
```
westonal: half a second delay? We have a lot of these, doesn't it just make the UI feel laggy.

I think in this case, you don't need RX, just an old onClickListener doing the thing immediately will be better experience.

ditn: Yes you're probably right, let me switch this over.

coreui/src/main/java/com/blockchain/ui/extensions/ThrottledClicks.kt
```diff
@@ -0,0 +1,11 @@
...
+
+import android.view.View
+import com.jakewharton.rxbinding2.view.clicks
+import io.reactivex.Observable
+import io.reactivex.android.schedulers.AndroidSchedulers
+import java.util.concurrent.TimeUnit
+
+fun View.throttledClicks(): Observable<Unit> =
+    clicks()
+        .throttleFirst(500, TimeUnit.MILLISECONDS, AndroidSchedulers.mainThread())
```
westonal: I'm interested in this scheduler choice, without it, do the clicks carry on on another thread?

ditn: Yes, unfortunately. Same as `debounce`:

```
@CheckReturnValue
@SchedulerSupport(SchedulerSupport.COMPUTATION)
public final Observable<T> throttleFirst(long windowDuration, TimeUnit unit) {
    return throttleFirst(windowDuration, unit, Schedulers.computation());
}
```

I thought about allowing the passing of a scheduler but it seemed a bit redundant.

PR #1153, opened 2018-09-04
 feat(HomeBrew): AND-1353 Nabu market service and mock webserver integration tests

westonal:
- I'm using the mock webserver from https://github.com/fabric8io/mockwebserver which I've previously used for websocket tests.
- I'm using koin to create my objects under test, and I only need to slightly override them in order to provide the mock webserver.
- The logic is really simple here, so I think end-to-end integration tests make sense, checking that with a given request and json response our model object comes out.


nabu/src/test/kotlin/com/blockchain/nabu/service/NabuMarketsServiceTest.kt
```diff
@@ -0,0 +1,80 @@
...
+        server.expect().get().withPath("/nabu-app/markets/quotes/BTC-ETH/config")
+            .andReturn(
+                200,
+                """
+{
+  "pair": "BTC-ETH",
+  "orderIncrement": "0.3",
+  "minOrderSize": "0.1",
+  "updated": "2018-08-28T15:42:04.490Z"
+}
```
westonal: Just one Json test, as they are ugly and hard to maintain, after this, in the next test I use a `TradingConfig` response. The mock server is serializing it for me.

nabu/src/test/kotlin/com/blockchain/nabu/service/apiServerTestModule.kt
```diff
@@ -0,0 +1,33 @@
...
+
+import com.blockchain.network.EnvironmentUrls
+import com.blockchain.network.modules.MoshiBuilderInterceptorList
+import com.blockchain.network.modules.OkHttpInterceptors
+import info.blockchain.balance.CryptoCurrency
+import io.fabric8.mockwebserver.DefaultMockServer
+import okhttp3.OkHttpClient
+import org.koin.dsl.module.applicationContext
+
+fun apiServerTestModule(server: DefaultMockServer) = applicationContext {
```
westonal: This module sets up the last needed bits for end-to-end testing and associates the mock server via it's base URL.

nabu/src/test/kotlin/com/blockchain/nabu/service/NabuMarketsServiceTest.kt
```diff
@@ -0,0 +1,80 @@
...
+import org.koin.test.AutoCloseKoinTest
+
+class NabuMarketsServiceTest : AutoCloseKoinTest() {
+
+    private val server = DefaultMockServer()
+
+    @get:Rule
+    val initMockServer = mockWebServerInit(server)
+
+    private val subject: NabuMarketsService by inject()
```
westonal: Notice I inject the SUT. So I have just one constructor call for this, in the Koin module.

nabu/src/test/kotlin/com/blockchain/nabu/service/NabuMarketsServiceTest.kt
```diff
@@ -0,0 +1,80 @@
...
+
+    @Before
+    fun startKoin() {
+        startKoin(
+            listOf(
+                apiModule,
+                nabuModule,
+                apiServerTestModule(server)
+            )
```
westonal: Couple of real modules, then an test module to complete the setup.

common/network/src/main/kotlin/com/blockchain/network/modules/apiModule.kt
```diff
@@ -100,4 +100,13 @@ val apiModule = applicationContext {
...
     }
+
+    bean("nabu") {
+        Retrofit.Builder()
+            .baseUrl(get<EnvironmentUrls>().apiUrl)
+            .client(get())
+            .addConverterFactory(get<MoshiConverterFactory>())
+            .addCallAdapterFactory(get<RxJava2CallAdapterFactory>())
+            .build()
+    }
```
westonal: This is what I think we should use instead of url concatenation.

nabu/src/main/kotlin/com/blockchain/nabu/api/NabuMarkets.kt
```diff
@@ -0,0 +1,13 @@
+package com.blockchain.nabu.api
+
+import io.reactivex.Single
+import retrofit2.http.GET
+import retrofit2.http.Path
+
+interface NabuMarkets {
+
+    @GET("nabu-app/markets/quotes/{trading_pair}/config")
```
westonal: Instead of URL concatenation.

nabu/src/main/kotlin/com/blockchain/koin/nabuModule.kt
```diff
@@ -0,0 +1,13 @@
+package com.blockchain.koin
+
+import com.blockchain.nabu.api.NabuMarkets
+import com.blockchain.nabu.service.NabuMarketsService
+import org.koin.dsl.module.applicationContext
+import retrofit2.Retrofit
+
+val nabuModule = applicationContext {
+
+    bean { get<Retrofit>("nabu").create(NabuMarkets::class.java) }
```
westonal: So, we can't use `"kotlin"` `Retrofit` instance because it's pointing to `explorerUrl`.
And we can't use `"api"` `Retrofit` instance because it's not a new moshi one.
I've gone for "nabu" as name, but I could also have named it `"api-kolin"` as nothing that nabu like about it.
This is the main reason you went for url concatenation I think, but I think it risky using a retrofit instance with a base url different to what you will actually hit.

nabu/src/main/kotlin/com/blockchain/nabu/service/NabuMarketsService.kt
```diff
@@ -0,0 +1,17 @@
...
+import info.blockchain.balance.CryptoValue
+import io.reactivex.Single
+
+class NabuMarketsService(private val nabuMarkets: NabuMarkets) {
+
+    fun getTradingConfig(tradingPair: CoinPair): Single<TradingConfig> {
+        return nabuMarkets.getTradingConfig(
+            tradingPair.pairCodeUpper
+        ).map {
+            TradingConfig(minOrderSize = CryptoValue.fromMajor(tradingPair.from, it.minOrderSize.toBigDecimal()))
```
westonal: Map it into something nice to work with in rest of program. Clear separation between DTO and model types.

nabu/src/main/kotlin/com/blockchain/nabu/api/NabuMarkets.kt
```diff
@@ -0,0 +1,13 @@
+package com.blockchain.nabu.api
+
+import io.reactivex.Single
+import retrofit2.http.GET
+import retrofit2.http.Path
+
+internal interface NabuMarkets {
```
westonal: As the Endpoints are `internal`, you must go though the service!

nabu/src/main/kotlin/com/blockchain/nabu/api/TradingConfig.kt
```diff
@@ -0,0 +1,5 @@
+package com.blockchain.nabu.api
+
+internal class TradingConfig(
```
westonal: DTO is internal, the backend's json format will be limited to this module.

nabu/src/main/kotlin/com/blockchain/nabu/service/NabuMarketsService.kt
```diff
@@ -0,0 +1,19 @@
+package com.blockchain.nabu.service
+
+import com.blockchain.morph.CoinPair
+import com.blockchain.nabu.api.NabuMarkets
+import info.blockchain.balance.CryptoValue
+import io.reactivex.Single
+
+class NabuMarketsService internal constructor(
```
westonal: `internal` constructor, you must create via the koin module defined in this module.

nabu/src/main/kotlin/com/blockchain/koin/nabuModule.kt
ditn: There were a few factors:

1. Initially the app supported dynamic switching of base URLs, albeit in a very clunky way. Switching to `@URL` would allow us to change things on the fly - but obviously if we're going to go the route of having different package names for different flavours, this is irrelevant.
2. Perceived cost of instantiating Retrofit. I think ultimately this probably isn't a big deal, and at the rate we're refactoring we can remove the legacy Jackson instances quite quickly anyway.
3. When initially setting up a Kotlin-specific instance, this gave me flexibility to use it across the API, Explorer, Coinify and other base URLs without having to manage each one (registering adapters etc).

But yeah, I agree this is the best approach going forwards. URL concatenation feels really nasty and unnecessary.

nabu/src/main/kotlin/com/blockchain/nabu/api/NabuMarkets.kt
ditn: Sensible.

nabu/src/main/kotlin/com/blockchain/nabu/api/TradingConfig.kt
ditn: Also sensible.

PR #1155, opened 2018-09-05
feat(HomeBrew): AND-1353 BigDecimal moshi adaptor and tests plus use in a DTO

westonal:


app/src/main/java/com/blockchain/koin/modules/moshiModule.kt
```diff
@@ -8,6 +8,7 @@ val moshiModule = applicationContext {
     bean {
         MoshiBuilderInterceptorList(
             listOf(
+                get("BigDecimal"),
```
ditn: Nitpicky, but PascalCase vs camelCase

ditn: Nah you're right actually, it's fine as-is.

PR #1157, opened 2018-09-06
feat(Homebrew): AND-1403 Users re-entering the KYC flow are now redir…

ditn:
…ected appropriately, and their data is restored throughout the entire fragment stack

### DIFF
--- a/kyc/src/main/java/com/blockchain/kycui/mobile/entry/KycMobileEntryPresenter.kt
+++ b/kyc/src/main/java/com/blockchain/kycui/mobile/entry/KycMobileEntryPresenter.kt
@@ -1,18 +1,26 @@
 package com.blockchain.kycui.mobile.entry
 
+import com.blockchain.kyc.datamanagers.nabu.NabuDataManager
+import com.blockchain.kycui.extensions.fetchNabuToken
 import com.blockchain.kycui.mobile.entry.models.PhoneDisplayModel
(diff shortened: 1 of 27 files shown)


kyc/src/main/java/com/blockchain/kycui/mobile/entry/KycMobileEntryPresenter.kt
```diff
@@ -25,7 +33,17 @@ class KycMobileEntryPresenter(
...
+                            nabuDataManager.requestJwt()
+                                .subscribeOn(Schedulers.io())
+                                .flatMap { jwt ->
+                                    fetchOfflineToken.flatMap {
+                                        nabuDataManager.updateUserWalletInfo(it, jwt)
+                                            .subscribeOn(Schedulers.io())
+                                    }
+                                }
+                                .ignoreElement()
+                        }
```
westonal: Two new dependencies to do just one thing is a smell.

It has also caused all this RX nesting.

It's currently too complicated to use authenticated endpoints, solution is to move to my `Authenticator`, and then it can be this simple:

```
.flatMapCompletable { _ ->
     nabuService.updateUserWalletInfo(it)
}
```

I say we add to tech debt AND-1363 as it's in now.

westonal: Added sub task to tech debt task AND-1410

PR #1160, opened 2018-09-07
feat(Homebrew): AND-1248 Redirect users according to region/KYC state…

ditn:
…/KYC account existence.

### DIFF
--- a/app/src/main/java/com/blockchain/koin/modules/morphMethodModule.kt
+++ b/app/src/main/java/com/blockchain/koin/modules/morphMethodModule.kt
@@ -1,18 +1,23 @@
 package com.blockchain.koin.modules
 
 import android.app.Activity
+import android.content.Intent
+import android.support.annotation.VisibleForTesting
 import com.blockchain.kycui.navhost.KycNavHostActivity
+import com.blockchain.kycui.settings.KycStatusHelper
+import com.blockchain.kycui.settings.SettingsKycState
 import com.blockchain.morph.MorphMethodSelector
+import com.blockchain.morph.ui.homebrew.exchange.ExchangeActivity
 import io.reactivex.Single
 import io.reactivex.android.schedulers.AndroidSchedulers
 import org.koin.dsl.module.applicationContext
-import piuk.blockchain.android.BuildConfig
 import piuk.blockchain.android.ui.shapeshift.overview.ShapeShiftActivity
 import timber.log.Timber
 
 enum class MorphMethodType {
     ShapeShift,
-    HomeBrew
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/blockchain/koin/modules/morphMethodModule.kt
```diff
@@ -1,18 +1,23 @@
...
 import org.koin.dsl.module.applicationContext
-import piuk.blockchain.android.BuildConfig
 import piuk.blockchain.android.ui.shapeshift.overview.ShapeShiftActivity
 import timber.log.Timber
 
 enum class MorphMethodType {
     ShapeShift,
-    HomeBrew
+    HomeBrew,
+    Kyc
```
ditn: I realise this wasn't really mean to include KYC, but this seemed the most sensible place to choose where exactly to send users as the KYC module doesn't have a dependency on Morph - therefore we can't redirect once a user has hit the KYC flow page which handles the rest of the routing.

app/src/main/java/com/blockchain/koin/modules/morphMethodModule.kt
```diff
@@ -53,15 +54,36 @@ val morphMethodModule = applicationContext {
...
+                                KycNavHostActivity.start(activity)
+                            }
                         }
                     }
         } as MorphActivityLauncher
     }
 }
 
-private fun fixedSelector(morphMethodType: MorphMethodType): MorphMethodTypeSelector =
+@VisibleForTesting
```
westonal: `internal` is visible for testing. You shouldn't need this. Also, now this can move to own file I feel.

app/src/main/java/com/blockchain/koin/modules/morphMethodModule.kt
westonal: Looks fine actually, nice simple solution. We might rename from `MorphMethodType` to something else. Can't think what though, so maybe that's fine too.

app/src/main/java/piuk/blockchain/android/ui/home/MainPresenter.java
```diff
@@ -195,35 +200,33 @@ void doTestnetCheck() {
...
-//        if (showShapeshift) {
-//            getView().showShapeshift();
-//        } else {
-//            getView().hideShapeshift();
-//        }
+
+        if (showExchange) {
+            getView().showExchange();
+        } else {
+            getView().hideExchange();
```
westonal: Funny how this wasn't used before (commented out).

PR #1168, opened 2018-09-12
feat(WebSocket): AND-1325 Enforce the marker interface JsonSerializable on Json Web Socket

westonal:
- Implementing JsonSerializable is an easy way to get proguard keeping and additional instance and class extension methods, so I'm enforcing it on the new Json Web Socket.

### DIFF
--- a/common/network/src/main/kotlin/com/blockchain/network/websocket/MoshiJsonWebSocketDecorator.kt
+++ b/common/network/src/main/kotlin/com/blockchain/network/websocket/MoshiJsonWebSocketDecorator.kt
@@ -1,12 +1,15 @@
 package com.blockchain.network.websocket
 
+import com.blockchain.serialization.JsonSerializable
 import com.squareup.moshi.JsonAdapter
 import com.squareup.moshi.Moshi
 import io.reactivex.Observable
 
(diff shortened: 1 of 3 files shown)


common/network/src/main/kotlin/com/blockchain/network/websocket/MoshiJsonWebSocketDecorator.kt
```diff
@@ -1,12 +1,15 @@
...
 import io.reactivex.Observable
 
-inline fun <reified OUTGOING : Any, reified INCOMING : Any> WebSocket<String, String>.toJsonSocket(
-    moshi: Moshi
-): WebSocket<OUTGOING, INCOMING> {
+inline fun <
+    reified OUTGOING : JsonSerializable,
+    reified INCOMING : JsonSerializable
+    >
+    WebSocket<String, String>.toJsonSocket(moshi: Moshi): WebSocket<OUTGOING, INCOMING> {
```
westonal: Bit messy, but this line was too long and I couldn't find any other indentation ktlint was happy with.

westonal:
Actually, the types you use can be manually converted to and from json and do not need this. `QuoteWebSocket` and it's Moshi adapters are an example of this. So I'm closing this.

PR #1172, opened 2018-09-12
feat(HomeBrew): AND-1339 Auth WebSocket

westonal:
Also:
- Timber extension for logging.


morph/homebrew/src/main/kotlin/com/blockchain/morph/homebrew/AuthWebSocket.kt
```diff
@@ -0,0 +1,37 @@
...
+                    AuthSubscribe(
+                        channel = "auth",
+                        operation = "subscribe",
+                        params = Params(
+                            type = "auth",
+                            token = it.token
+                        )
+                    ).toMoshiJson()
+                )
+            }
```
westonal: After open this sends the authenticate message. It doesn't look for the response.

PR #1174, opened 2018-09-13
feat(Homebrew): AND-1372 Add execution API endpoint + objects

ditn:


nabu/src/main/kotlin/com/blockchain/nabu/api/TradeRequest.kt
```diff
@@ -0,0 +1,36 @@
...
+    val destinationAddress: String,
+    val refundAddress: String,
+    val quote: ReceivedQuote
+) : JsonSerializable
+
+class ReceivedQuote(
+    val pair: String,
+    val fiatCurrency: String,
+    val fix: String,
+    val volume: Double,
```
westonal: Probably want to use `BigDecimal` here.

nabu/src/main/kotlin/com/blockchain/nabu/api/TradeRequest.kt
```diff
@@ -0,0 +1,36 @@
...
+) : JsonSerializable
+
+class CryptoAndFiat(
+    val fiat: Value,
+    val crypto: Value
+) : JsonSerializable
+
+class Value(
+    val symbol: String,
+    val value: Double
```
westonal: Probably want to use `BigDecimal` here.

PR #1178, opened 2018-09-14
feat(HomeBrew): AND-1350 Live quotes

westonal:


ditn:
`ExchangeActivityConfigurationChangePersistenceTest` is not building unfortunately.

westonal:
The one time I don't run tests locally before pushing...

PR #1179, opened 2018-09-16
feat(HomeBrew): AND-1352 Adding mock data to test the layouts

robertogdsbc:
feat(HomeBrew): AND-1352 Exchange screens
- Added mock data to test the layouts.
- Also added the adapter for trade history recyclerView.
- TODO: its just mock data so it has to be connected with the real endpoint and use the morphTrade interface to use different trade models This is using a simple Trade model.


morph/ui/src/main/java/com/blockchain/morph/ui/homebrew/exchange/ExchangeConfirmationActivity.kt
```diff
@@ -2,24 +2,18 @@ package com.blockchain.morph.ui.homebrew.exchange
...
-    private lateinit var toolbar: Toolbar
-
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         setContentView(R.layout.activity_homebrew_trade_confirmation)
 
-        sendButton = findViewById(R.id.button_send_now)
-
-        sendButton.setOnClickListener {
+        button_send_now.setOnClickListener {
```
ditn: Why have we switched here? We're not using kotlinx in Morph, see conversation in Slack.

morph/ui/src/main/java/com/blockchain/morph/ui/homebrew/exchange/detail/HomebrewTradeDetailActivity.kt
```diff
@@ -0,0 +1,28 @@
...
+
+class HomebrewTradeDetailActivity : BaseAuthActivity() {
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        setContentView(R.layout.activity_homebrew_trade_detail)
+
+        val trade = intent.extras.get("EXTRA_TRADE") as Trade
+        setupToolbar(R.id.toolbar_constraint, R.string.exchange)
+        supportActionBar?.title = trade.id
```
ditn: Here you're setting the Toolbar title twice..?

robertogdsbc: afaik setupToolbar params are (Int, Int) didn't want to touch it but I can override setupToolbar to support String

ditn: Yeah please do. Otherwise the font won't be applied I think.

ditn:
Conflicts.

PR #1180, opened 2018-09-17
 feat(HomeBrew): AND-1350 Live quotes - part 2

westonal:
- Move classes and tidy dependencies
- Had to add an abstraction layer to logging to allow Timber calls from pure Kt modules, like network


common/network/src/test/kotlin/com/blockchain/network/websocket/LoggingWebSocketTest.kt
```diff
@@ -0,0 +1,118 @@
...
+                    }
+                }
+            )
+        )
+    }
+
+    @Test
+    fun `no wrapper socket is provided`() {
+        val underlyingSocket = mock<WebSocket<String, String>>()
+        underlyingSocket.debugLog("X") `should be` underlyingSocket
```
westonal: So `debugLog` is guaranteed to do nothing and have no runtime overhead when the `NullLogger` is registered.

morph/ui/build.gradle
```diff
@@ -35,12 +35,6 @@ dependencies {
     api project(':morph:common')
     implementation project(':balance')
 
-    // TODO: AND-1363 Temporary - need to move much of the logic from ExchangeActivity to :morph:homebrew and place
-    // generic interfaces in :morph:common, then these can go
-    api project(':common:network')
-    api project(':morph:homebrew')
-    api project(':nabu')
```
westonal: Was able to clear all of these, the exchange activity now uses only generic `:common:morph` references.

PR #1181, opened 2018-09-17
chore(Transactions): AND-1372 Create single datamanager for handling all cryptocurrency transactions

ditn:
All of this code is repeated in a few different places now; this puts it in one singular place.

More than happy for suggestions if you've got any bright ideas on how to simplify this further.


core/src/main/java/com/blockchain/datamanagers/TransactionSendDataManager.kt
```diff
@@ -0,0 +1,204 @@
...
+                                amount,
+                                spendable,
+                                signingKeys,
+                                destination,
+                                it,
+                                feePerKb
+                            )
+                        }
+                }
+        }
```
westonal: Looks like only difference between this and the previous method is the account. Maybe extract a function:

```
private fun sendBtcTransaction(
        amount: CryptoValue,
        destination: String,
        account: Account,
        feePerKb: BigInteger
    ): Single<String> = sendBitcoinStyleTransaction(these params,  account)

private fun sendBchTransaction(
        amount: CryptoValue,
        destination: String,
        account: GenericMetadataAccount,
        feePerKb: BigInteger
    ): Single<String> = sendBitcoinStyleTransaction(these params,  account.getHdAccount())
```

core/src/test/java/com/blockchain/datamanagers/TransactionSendDataManagerTest.kt
```diff
@@ -0,0 +1,319 @@
...
+            bchDataManager,
+            sendDataManager,
+            ethereumAccountWrapper
+        )
+    }
+
+    @Test
+    fun `execute bitcoin transaction should attempt to send bitcoin`() {
+        // Arrange
+        val amount = CryptoValue(CryptoCurrency.BTC, BigInteger.TEN)
```
westonal: In test utils there is:
```
10.0.bitcoin()
```
as a neat way of writing `10 BTC`, or:
```
CryptoValue.bitcoinFromSatoshis(10L)
```
For 10 satoshis.

Using the constructor has given you satoshis, and I don't know you intended that.

core/src/test/java/com/blockchain/datamanagers/TransactionSendDataManagerTest.kt
```diff
@@ -0,0 +1,319 @@
...
+        subject.executeTransaction(amount, destination, account, feeOptions)
+            .test()
+        // Assert
+        verify(sendDataManager).getUnspentBchOutputs("XPUB")
+    }
+
+    @Test
+    fun `execute ethereum transaction should attempt to send ethereum`() {
+        // Arrange
+        val amount = CryptoValue(CryptoCurrency.ETHER, BigInteger.TEN)
```
westonal: `10.0.ether()`

core/src/test/java/com/blockchain/datamanagers/TransactionSendDataManagerTest.kt
```diff
@@ -0,0 +1,319 @@
...
+        val amount = CryptoValue(CryptoCurrency.BCH, BigInteger.TEN)
+        val destination = "DESTINATION"
+        val account = GenericMetadataAccount().apply { xpub = "XPUB" }
+        whenever(sendDataManager.getUnspentBchOutputs("XPUB"))
+            .thenReturn(Observable.empty())
+        // Act
+        subject.executeTransaction(amount, destination, account, feeOptions)
+            .test()
+        // Assert
+        verify(sendDataManager).getUnspentBchOutputs("XPUB")
```
westonal: This isn't testing that the correct transaction was sent, just that some method was called.

I don't think these tests have any value, they just make the tests rigid.

core/src/test/java/com/blockchain/datamanagers/TransactionSendDataManagerTest.kt
```diff
@@ -0,0 +1,319 @@
...
+            feeOptions.regularFee.toBigInteger()
+        )
+        verify(sendDataManager).submitBtcPayment(
+            spendable,
+            listOf(ecKey),
+            destination,
+            change,
+            feeOptions.regularFee.toBigInteger(),
+            amount.amount
+        )
```
westonal: This is a much better test, because it verifies the only important thing, that the correct transaction was sent.

However, it verifies things it doesn't need to:
```
        verify(sendDataManager).getUnspentOutputs("XPUB")
        verify(sendDataManager).getSpendableCoins(
            unspentOutputs,
            BigInteger.TEN,
            feeOptions.regularFee.toBigInteger()
        )
```
These verifies are not needed, if they weren't called, `submitBtcPayment` wouldn't have the correct values. That's also why the other tests that only have these have little value.

core/src/test/java/com/blockchain/datamanagers/TransactionSendDataManagerTest.kt
```diff
@@ -0,0 +1,319 @@
...
+        verify(ethDataManager).pushEthTx(signedTx)
+        verify(ethDataManager).setLastTxHashObservable(eq(txHash), any())
+    }
+
+    private val feeOptions = FeeOptions().apply {
+        priorityFee = 100L
+        regularFee = 10L
+        gasLimit = 21000L
+    }
```
westonal: Can we look at making `FeeOptions` immutable?

core/src/test/java/com/blockchain/datamanagers/TransactionSendDataManagerTest.kt
westonal: I will look at making the constructor private. It usually reads better using one of the static factory methods.

core/src/test/java/com/blockchain/datamanagers/TransactionSendDataManagerTest.kt
ditn: Hmm this is used in dozens of places, so I might skip this for now.

westonal:
Some thing's wrong with the rebase, I see my commit in the commit list.

image

The "rebase" commit here is actually a merge, rebase shouldn't generate any extra commits.

ditn:
That's very odd. Maybe I'll re-open this one.

PR #1187, opened 2018-09-19
feat(HomeBrew): AND-1350 Add Nabu url to secrets

westonal:


app/src/main/java/com/blockchain/koin/modules/nabuUrlModule.kt
```diff
@@ -0,0 +1,14 @@
...
+
+import com.blockchain.network.websocket.Options
+import org.koin.dsl.module.applicationContext
+
+val nabuUrlModule = applicationContext {
+
+    bean("nabu") {
+        Options(
+            name = "Nabu",
+            url = getProperty(WebSocketUrlKeys.Nabu)
```
westonal: Maybe it's all a bit OTT. I can just reference `BuildConfig` here directly.

westonal: Yeah I've pushed new commit to simplify.

PR #1188, opened 2018-09-19
feat(HomeBrew): AND-1350 Very basic fake quote service for demo app as real service requires auth

westonal:


morph/morphdevapp/src/main/java/com/blockchain/morph/dev/FakeQuoteService.kt
```diff
@@ -0,0 +1,70 @@
...
+            }
+        )
+    }
+
+    override fun openAsDisposable(): Disposable {
+        return CompositeDisposable()
+    }
+
+    override val quotes: Observable<Quote>
+        get() = subject.delay(1, TimeUnit.SECONDS)
```
westonal: Complete with fake delay. Actually exposes some of the funky behaviour I see occasionally with the real one.

PR #1189, opened 2018-09-19
feat(Homebrew): AND-1352 Add order history endpoint, render in UI

ditn:
screenshot1537365061


app/src/main/res/values/strings.xml
```diff
@@ -21,8 +21,6 @@
     <string name="polite_no">No thanks</string>
     <string name="done">Done</string>
 
-    <string name="YESTERDAY">Yesterday</string>
-
     <string name="SENT">Sent</string>
     <string name="sending">Sending</string>
     <string name="RECEIVED">Received</string>
```
robertogdsbc: just curious why we have this uppercase ids.

ditn: Legacy strings from 3-4 years ago.

PR #1193, opened 2018-09-20
feat(Homebrew): AND-1386 Render real data on order detail page

ditn:
screenshot1537440030

### DIFF
--- a/morph/ui/src/main/java/com/blockchain/morph/ui/homebrew/exchange/detail/HomebrewTradeDetailActivity.kt
+++ b/morph/ui/src/main/java/com/blockchain/morph/ui/homebrew/exchange/detail/HomebrewTradeDetailActivity.kt
@@ -1,6 +1,5 @@
 package com.blockchain.morph.ui.homebrew.exchange.detail
 
-import android.annotation.SuppressLint
 import android.content.ClipData
 import android.content.ClipboardManager
 import android.content.Context
@@ -27,7 +26,6 @@ import piuk.blockchain.androidcoreui.utils.extensions.visible
 
 class HomebrewTradeDetailActivity : BaseAuthActivity() {
 
-    @SuppressLint("SetTextI18n")
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         setContentView(R.layout.activity_homebrew_trade_detail)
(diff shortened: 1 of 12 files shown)


morph/ui/src/main/java/com/blockchain/morph/ui/homebrew/exchange/detail/HomebrewTradeDetailActivity.kt
```diff
@@ -1,29 +1,160 @@
...
-        // TODO: data to test layout - use real object
         status.text = trade.state.toStatusString(this)
-        value.text = trade.depositQuantity
+        status.setCompoundDrawablesWithIntrinsicBounds(
+            trade.state.toDrawable(this),
+            null,
+            null,
+            null
+        )
+        value.text = if (trade.approximateValue()) "~${trade.price}" else trade.price
```
westonal: Does this need localization?
Does `~` mean approx. in every locale?

morph/ui/src/main/java/com/blockchain/morph/ui/homebrew/exchange/detail/HomebrewTradeDetailActivity.kt
```diff
@@ -1,29 +1,160 @@
...
-        fees.text = trade.price
+        fees.text = trade.fee
+
+        if (trade.shouldStrike()) {
+            receive.paintFlags = Paint.STRIKE_THRU_TEXT_FLAG or Paint.ANTI_ALIAS_FLAG
+        }
+
+        if (trade.refunding()) {
+            receive.text = "~${trade.quantity}"
+            fees.text = "~${trade.fee}"
```
westonal: Whether localizable or not, I would at least extract a function for this `trade.fee.displayAsApproximate()`

morph/ui/src/main/java/com/blockchain/morph/ui/homebrew/exchange/detail/HomebrewTradeDetailActivity.kt
ditn: That's a really good question. It's a mathematical symbol so I would assume so but let me check.

ditn: Doesn't look like it's something that's localised anywhere.

PR #1200, opened 2018-09-24
feat(Homebrew): AND-1435 Fixes for trade execution - whole flow shoul…

ditn:
…d be working now

### DIFF
--- a/nabu/src/main/kotlin/com/blockchain/nabu/service/NabuMarketsService.kt
+++ b/nabu/src/main/kotlin/com/blockchain/nabu/service/NabuMarketsService.kt
@@ -83,7 +83,7 @@ private fun TradeJson.map(): NabuTransaction {
         id = this.id,
         createdAt = this.createdAt,
         pair = coinPair,
-        rate = this.rate,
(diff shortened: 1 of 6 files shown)


nabu/src/main/kotlin/com/blockchain/nabu/service/NabuMarketsService.kt
```diff
@@ -83,7 +83,7 @@ private fun TradeJson.map(): NabuTransaction {
         id = this.id,
         createdAt = this.createdAt,
         pair = coinPair,
-        rate = this.rate,
+        rate = this.rate ?: withdrawal.value / deposit.value,
```
westonal: If this is all it is, I don't know why backend would ever have to give us `null`?

ditn: I assume we're supposed to render the rate as approximate, but then we also never render the rate so I guess this doesn't matter at all.

westonal: I see. Well even if we are supposed to render an approximate, backend can provide that approximate.

PR #1201, opened 2018-09-24
feat(HomeBrew): AND-1350 Keyboard Crypto Entry

westonal:
- The Crypto entry works differently to fiat
- There are still issues around this but this is much better than the version in develop
- Issues include:
- Doesn't resize font of value is getting large
- Doesn't always update where there are trailing zeros

### DIFF
--- a/balance/src/main/kotlin/info/blockchain/balance/Money.kt
+++ b/balance/src/main/kotlin/info/blockchain/balance/Money.kt
@@ -1,5 +1,6 @@
 package info.blockchain.balance
 
+import java.math.BigDecimal
 import java.util.Locale
 
 interface Money {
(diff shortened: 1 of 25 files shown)


balance/src/main/kotlin/info/blockchain/balance/Money.kt
```diff
@@ -1,8 +1,69 @@
...
+            }
+        }
+
+    class Parts(
+        val symbol: String,
+        val major: String,
+        val minor: String,
+        val majorAndMinor: String
+    )
+}
```
westonal: I've unified a lot of behaviour and logic into `Money`.

ditn: Yeah looks good.

PR #1204, opened 2018-09-24
feat(HomeBrew): AND-1350 Disable Exchange button when a quote is not valid

westonal:
- Button is disabled and grey when the last quote doesn't match the value the user has typed.
- This happens when the quote has not yet been returned
- Or the quote was unable to be returned for whatever reason

image

### DIFF
--- a/morph/ui/src/main/res/layout/fragment_homebrew_exchange.xml
+++ b/morph/ui/src/main/res/layout/fragment_homebrew_exchange.xml
@@ -149,7 +149,6 @@
         android:layout_marginBottom="16dp"
         android:layout_marginEnd="16dp"
         android:layout_marginStart="16dp"
-        android:background="@drawable/exchange_button_background"
         app:layout_constraintBottom_toBottomOf="parent"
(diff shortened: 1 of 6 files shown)


morph/ui/src/main/res/layout/fragment_homebrew_exchange.xml
```diff
@@ -149,7 +149,7 @@
         android:layout_marginBottom="16dp"
         android:layout_marginEnd="16dp"
         android:layout_marginStart="16dp"
-        android:background="#FF10ADE4"
+        android:background="@drawable/exchange_button_background"
```
ditn: You shouldn't really need to define the button's background this way, as you won't get proper theming and button states. `android:theme="@style/LightBlueButtonTheme"` should do everything you need.

westonal: I'll give it a go

westonal: image

PR #1206, opened 2018-09-25
feat(Homebrew): AND-1319 Allow country search via acronym, prioritise…

ditn:
… such results


kyc/src/main/java/com/blockchain/kycui/countryselection/util/NabuCountryResponseListExtensions.kt
```diff
@@ -24,4 +24,10 @@ data class CountryDisplayModel(
...
     val flag: String
-)
\ No newline at end of file
+) {
+    val searchCode = "${name.acronym()};$countryCode;$name"
+}
+
+internal fun String.acronym(): String = this.toCharArray()
+    .filter { it.isUpperCase() }
+    .joinToString(separator = "")
```
westonal: Other option to `joinToString`:
```internal fun String.acronym(): String = String(this.toCharArray().filter { it.isUpperCase() }.toCharArray())
```

ditn: Ah that would have been a lot neater.

PR #1208, opened 2018-09-25
fix(Homebrew): AND-1421 Fix issue where click events wouldn't registe…

ditn:
…r after returning to the page


kyc/src/main/java/com/blockchain/kycui/settings/KycStatusHelper.kt
```diff
@@ -89,9 +89,11 @@ class KycStatusHelper(
             .map { it.countryCode }
             .singleOrError(),
         BiFunction { countries: List<NabuCountryResponse>, countryCode: String ->
-            countries
-                .map { it.code }
-                .contains(countryCode)
+            // STOPSHIP: For testing purposes only
```
westonal: Supposed to be in the PR?

PR #1214, opened 2018-09-26
fix(Homebrew): AND-1467 Withdrawal can be null - handle this possibility

ditn:


ditn:
@westonal This can be reviewed, it's only marked as ❌ due to Travis

nabu/src/main/kotlin/com/blockchain/nabu/service/NabuMarketsService.kt
```diff
@@ -82,18 +84,20 @@ private fun PeriodicLimit.toFiat(currencyCode: String) =
...
-        rate = this.rate ?: withdrawal.value / deposit.value,
+        rate = rate,
         fee = this.withdrawalFee.toCryptoValue(),
         fiatValue = this.fiatValue.toFiatValue(),
         refundAddress = this.refundAddress,
         depositAddress = this.depositAddress,
         deposit = this.deposit.toCryptoValue(),
         withdrawalAddress = this.withdrawalAddress,
-        withdrawal = this.withdrawal.toCryptoValue(),
+        withdrawal = this.withdrawal?.toCryptoValue() ?: CryptoValue(coinPair.to, BigInteger.ZERO),
```
westonal: `CryptoValue.zero(coinPair.to)` - doesn't create a new instance that's all.

ditn: Ah yeah nice, I'll change this.

PR #1219, opened 2018-09-26
feat(Homebrew): AND-1440 Request trade limits in user's currency choice

westonal:
- In addition, fixed issue with how KYC reported the users currency during sign up.

I don't know if these together will solve AND-1440, but they may.

### DIFF
--- a/app/src/main/java/piuk/blockchain/android/ui/home/MainActivity.java
+++ b/app/src/main/java/piuk/blockchain/android/ui/home/MainActivity.java
@@ -494,7 +494,7 @@ public class MainActivity extends BaseMvpActivity<MainView, MainPresenter> imple
                 MorphMethodModuleKt.launchAsync(morphActivityLauncher, MainActivity.this);
                 break;
             case R.id.nav_exchange_homebrew_debug:
-                HomebrewNavHostActivity.start(this, "GBP");
+                HomebrewNavHostActivity.start(this, mainPresenter.getDefaultCurrency());
                 break;
             case R.id.nav_addresses:
                 startActivityForResult(new Intent(this, AccountActivity.class), ACCOUNT_EDIT);
--- a/app/src/main/java/piuk/blockchain/android/ui/settings/SettingsPresenter.java
+++ b/app/src/main/java/piuk/blockchain/android/ui/settings/SettingsPresenter.java
@@ -116,7 +116,7 @@ public class SettingsPresenter extends BasePresenter<SettingsView> {
                         .subscribe(
                                 settingsKycState -> {
                                     if (settingsKycState == SettingsKycState.Verified) {
-                                        getView().launchHomebrew(prefsUtil.getValue(PrefsUtil.DEFAULT_CURRENCY, PrefsUtil.DEFAULT_CURRENCY));
+                                        getView().launchHomebrew(prefsUtil.getValue(PrefsUtil.KEY_SELECTED_FIAT, PrefsUtil.DEFAULT_CURRENCY));
                                     } else if (settingsKycState == SettingsKycState.Unverified) {
                                         getView().launchKycFlow();
                                     } else {
--- a/app/src/main/java/piuk/blockchain/android/ui/settings/SettingsPresenter.java
+++ b/app/src/main/java/piuk/blockchain/android/ui/settings/SettingsPresenter.java
@@ -116,7 +116,7 @@ public class SettingsPresenter extends BasePresenter<SettingsView> {
                         .subscribe(
                                 settingsKycState -> {
                                     if (settingsKycState == SettingsKycState.Verified) {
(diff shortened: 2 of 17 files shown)


app/src/main/java/piuk/blockchain/android/ui/home/MainActivity.java
```diff
@@ -494,7 +494,7 @@ public void selectDrawerItem(MenuItem menuItem) {
                 MorphMethodModuleKt.launchAsync(morphActivityLauncher, MainActivity.this);
                 break;
             case R.id.nav_exchange_homebrew_debug:
-                HomebrewNavHostActivity.start(this, "GBP");
+                HomebrewNavHostActivity.start(this, mainPresenter.getDefaultCurrency());
```
westonal: I liked this, debug should use the users currency too. Makes things easier.

app/src/main/java/piuk/blockchain/android/ui/settings/SettingsPresenter.java
```diff
@@ -116,7 +116,7 @@ void onKycStatusClicked() {
                         .subscribe(
                                 settingsKycState -> {
                                     if (settingsKycState == SettingsKycState.Verified) {
-                                        getView().launchHomebrew(prefsUtil.getValue(PrefsUtil.DEFAULT_CURRENCY, "USD"));
+                                        getView().launchHomebrew(prefsUtil.getValue(PrefsUtil.KEY_SELECTED_FIAT, PrefsUtil.DEFAULT_CURRENCY));
```
westonal: This is the KYC issue. Always reported USD for a user.

core/src/main/java/piuk/blockchain/androidcore/utils/FiatCurrencyPreference.kt
```diff
@@ -0,0 +1,6 @@
+package piuk.blockchain.androidcore.utils
+
+interface FiatCurrencyPreference {
+
+    val fiatCurrencyPreference: String
```
westonal: Instead of having to pass keys and defaults around, we should move to using interfaces like this and a single instance of the `getValue` call.

morph/homebrew/src/main/kotlin/com/blockchain/morph/homebrew/ExchangeWebSocket.kt
```diff
@@ -37,9 +37,9 @@ class ExchangeWebSocket(private val underlyingSocket: WebSocket<String, String>,
                         val split = rate.pair.split("-")
                         val lhs = split.firstOrNull()
                         val rhs = split.lastOrNull()
-                        val fromCrypto = CryptoCurrency.fromSymbol(rhs)
-                        if (fromCrypto != null && lhs != null) {
-                            ExchangeRate.CryptoToFiat(fromCrypto, lhs, rate.price)
+                        val fromCrypto = CryptoCurrency.fromSymbol(lhs)
+                        if (fromCrypto != null && rhs != null) {
+                            ExchangeRate.CryptoToFiat(fromCrypto, rhs, rate.price)
```
westonal: They had a bug on backend, flipped symbol. This means that the released version today will not see the actual users maximum as it won't see the exchange rate.

app/src/main/java/piuk/blockchain/android/ui/home/MainActivity.java
ditn: Hah okay cool.

app/src/main/java/piuk/blockchain/android/ui/settings/SettingsPresenter.java
ditn: Oh nicely spotted. My mistake.

morph/homebrew/src/main/kotlin/com/blockchain/morph/homebrew/ExchangeWebSocket.kt
ditn: Oh that's a shame. Maybe I could sneak in another build this morning.

morph/ui/src/main/java/com/blockchain/morph/ui/homebrew/exchange/history/TradeHistoryActivity.kt
```diff
@@ -27,7 +27,7 @@ class TradeHistoryActivity : BaseMvpActivity<TradeHistoryView, TradeHistoryPrese
 
     override val locale: Locale = Locale.getDefault()
     private val presenter: TradeHistoryPresenter by inject()
-    private val fiat: CurrencyState by inject()
+    private val fiat: FiatCurrencyPreference by inject()
```
ditn: Yeah this is cool.

PR #1221, opened 2018-09-27
feat(Homebrew): AND-1457 Keep fiat entry on account change

westonal:


morph/ui/src/main/java/com/blockchain/morph/ui/homebrew/exchange/ExchangeModel.kt
```diff
@@ -56,8 +66,28 @@ class ExchangeModel(
         Timber.d("ExchangeViewModel cleared")
     }
 
-    fun newDialog(
+    private fun initDialog(quoteService: QuoteService) {
+        val fiatCurrency = currencyPreference.fiatCurrencyPreference
+        newDialog(
+            fiatCurrency,
+            quoteService,
+            ExchangeDialog(
```
westonal: There is now just one MVI dialog for the lifespan of the exchange, there is no need for the "configChangePersistence" I had which stored the last values for restoring.
Note, the socket still gets closed on pause and reconnected on resume.

PR #1223, opened 2018-09-27
fix(Metadata): AND-909 Ensure that the Koin objects are being used rather than having Dagger recreate them too. Migrate some schedulers and update some logging.

ditn:
I was consistently seeing errors fetching metadata on first login when creating a new wallet, due to the wallet body being empty - which didn't make any sense. Here I've ensured that we're using Koin to construct all of the related classes and now I'm not seeing any issues.


westonal:
Interesting, I've only seen this issue before with `@Singleton`s, but makes sense how it can happen, good find.

PR #1225, opened 2018-09-27
feat(Homebrew): AND-1442 Resize font when gets long, also limit entry

westonal:
- to 11 significant figures or 999,999 whatever comes first

image

image

### DIFF
--- a/morph/ui/src/main/java/com/blockchain/morph/ui/homebrew/exchange/ExchangeFragment.kt
+++ b/morph/ui/src/main/java/com/blockchain/morph/ui/homebrew/exchange/ExchangeFragment.kt
@@ -1,9 +1,12 @@
 package com.blockchain.morph.ui.homebrew.exchange
 
+import android.annotation.SuppressLint
 import android.content.Context
 import android.os.Bundle
 import android.support.annotation.IdRes
 import android.support.v4.app.Fragment
+import android.support.v4.widget.TextViewCompat
+import android.util.TypedValue
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
@@ -17,6 +20,7 @@ import com.blockchain.morph.exchange.mvi.ApplyMinimumLimit
 import com.blockchain.morph.exchange.mvi.ExchangeIntent
 import com.blockchain.morph.exchange.mvi.ExchangeViewState
 import com.blockchain.morph.exchange.mvi.Fix
+import com.blockchain.morph.exchange.mvi.Maximums
 import com.blockchain.morph.exchange.mvi.SimpleFieldUpdateIntent
 import com.blockchain.morph.exchange.mvi.SwapIntent
 import com.blockchain.morph.exchange.mvi.ToggleFiatCryptoIntent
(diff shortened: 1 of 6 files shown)


morph/ui/src/main/java/com/blockchain/morph/ui/homebrew/exchange/ExchangeFragment.kt
```diff
@@ -131,6 +142,13 @@ internal class ExchangeFragment : Fragment() {
     override fun onResume() {
         super.onResume()
 
+        keyboard.setMaximums(
+            Maximums(
+                maxDigits = 11,
+                maxValue = 999_999.toBigDecimal()
```
westonal: The maximums for the keyboard are very flexible if we need different upper limits. I thought 11 significant digits and nearly 1 million was enough for all currencies.

morph/ui/src/main/java/com/blockchain/morph/ui/homebrew/exchange/ExchangeFragment.kt
```diff
@@ -201,24 +219,42 @@ internal class ExchangeFragment : Fragment() {
...
 
         val fromFiatString = value.fiatValue.toStringWithSymbol()
         smallValue.text = fromFiatString
     }
 
+    private fun setLargeText(largeText: String) {
+        largeValue.setTextSize(
+            TypedValue.COMPLEX_UNIT_PX,
+            resources.getDimension(
+                if (largeText.length > 8) R.dimen.large_exchange_text_size_long else R.dimen.large_exchange_text_size
```
westonal: Two sizes are used for fiat, while crypto gets a full resizing to fix experience. This is because I can't have the symbols floating around like they do for fiat and use the full resizing as I need `match_parent` width for it.

morph/ui/src/main/java/com/blockchain/morph/ui/homebrew/exchange/ExchangeFragment.kt
ditn: Yeah I'm sure this won't be a problem in the real world.

morph/ui/src/main/java/com/blockchain/morph/ui/homebrew/exchange/ExchangeFragment.kt
ditn: We could potentially use 

```
app:autoSizePresetSizes="@array/autosize_text_sizes"
app:autoSizeTextType="uniform"
```

instead of having to handle this manually and define an array of supported sizes. Depends how predictable you want the behaviour to be I guess.

westonal: I can't have `match_parent` on this textview, so autoSizeText doesn't work :(

ditn: Ah no, fair enough.

PR #1229, opened 2018-09-28
fix(Homebrew): AND-1483 Show ellipsis on end of order ID, fix issue w…

ditn:
…ith disclaimer text


ditn:
Maybe yeah. The issue is on very small screens that the text could be very very small. Elsewhere we've used an animation to expose the whole ID but I might check with Rosana today what she thinks is best.

PR #1233, opened 2018-09-30
chore(Build): AND-1486 Move zxing

westonal:
- Xzing related code is over 9% of the source files in `:app` 33/354.
- Made new module


zxing/src/main/java/piuk/blockchain/android/ui/zxing/CaptureActivity.java
```diff
@@ -389,15 +389,15 @@ public void drawViewfinder() {
...
-            case R.id.action_flash_light:
-                try {
-                    toggleTorch();
-                } catch (Exception e) {
-                }
-                return true;
-            default:
-                return super.onOptionsItemSelected(item);
+        int i = item.getItemId();
+        if (i == R.id.action_flash_light) {
```
westonal: In library modules, ids are not static, so can't be used in switch cases. Used android studio to convert it.

zxing/src/main/java/piuk/blockchain/android/ui/zxing/CaptureActivityHandler.java
```diff
@@ -78,58 +78,56 @@
...
+        intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
+        intent.putExtra(Browser.EXTRA_APPLICATION_ID, browserPackageName);
+      }
+
+      try {
+        activity.startActivity(intent);
+      } catch (ActivityNotFoundException anfe) {
+        Log.w(TAG, "Can't find anything to handle VIEW of URI " + url);
+      }
+
```
westonal: Another id in case statement issue.

zxing/src/main/java/piuk/blockchain/android/ui/zxing/CaptureActivity.java
ditn: Interesting! I didn't realise this.

PR #1237, opened 2018-10-01
fix(Homebrew): AND-1473 Allow users who select a valid SS country to …

ditn:
…be redirected to SS until the country becomes available in KYC


kyc/src/main/java/com/blockchain/kycui/countryselection/KycCountrySelectionPresenter.kt
```diff
@@ -42,13 +47,28 @@ class KycCountrySelectionPresenter(
...
     }
 
+    private fun checkShapeShift(countryCode: String) {
+        compositeDisposable +=
+            walletOptionsDataManager.isInShapeShiftCountry(countryCode)
+                .subscribeBy(
+                    onSuccess = {
+                        if (it) {
+                            prefsUtil.setValue(PREF_CHOSEN_KYC_COUNTRY, countryCode)
+                            view.redirectToShapeShift()
```
westonal: This happens as soon as country is selected? Sounds too easy to make a mistake and get stuck.

PR #1240, opened 2018-10-02
fix(Homebrew): AND-1516 Handle Nabu failure on login more gracefully

ditn:


app/src/main/java/piuk/blockchain/android/ui/home/MainPresenter.java
```diff
@@ -211,12 +211,10 @@ private void doWalletOptionsChecks() {
                 (showShapeShift, showKyc) -> showShapeShift || showKyc
         ).observeOn(AndroidSchedulers.mainThread())
                 .compose(RxUtil.addSingleToCompositeDisposable(this))
-                .subscribe(this::setExchangeVisiblity, throwable -> {
-                    //Couldn't retrieve wallet options. Not safe to continue
```
ditn: This is called in a few more places now, including wallet startup - so this warning isn't really relevant anymore.

app/src/main/java/piuk/blockchain/android/ui/home/MainPresenter.java
```diff
@@ -211,12 +211,10 @@ private void doWalletOptionsChecks() {
...
-                    Timber.e(throwable);
-                    getView().showToast(R.string.unexpected_error, ToastCustom.TYPE_ERROR);
-                    accessState.logout(applicationContext);
```
westonal: You needed to take this out as well?

PR #1247, opened 2019-01-31
Update strings.xml

sui77:
(native german speaker here, who just googled that error message cause it made no sense...) 

invalid amount = Ungültiger Betrag
invalid account = Ungültiges Konto


westonal:
Thank you very much!
We manage strings using an external service, so I will be sure to update this via that.

westonal:
Can you fix this string please @robertogdsbc ?

PR #1295, opened 2021-07-20
Fixed scanning for ETH & BTC (with amount of crypto in QR)

alexsoukhov:
## Objective

Fixed QR code scanning with amount/value fields provided for BTC/ETH.

## Description

Changed parseAddress methods of EthAsset and BtcAsset classes.

## How to Test

Scan QR code with value/amount in it. Like:
ethereum:0x5F36F007EDBE4218C3BB7CAD085DB7945E84C5E4?value=169950000000000000000&gas=25000
bitcoin:bc1q9wkggcz3s5npwehudhddkkxmmfkcsg6rlaa3jn?amount=0.0000119

## Screenshot/Design assessment

*If the change is UI related, paste a screenshot here and include an assessment of design accuracy if applicable.*

## Merge Checklist

- [ ] The PR uses a title supported by [.changelogrc](https://github.com/blockchain/My-Wallet-V3-iOS/blob/dev/.changelogrc#L6...L69).
- [ ] You have added unit tests.
- [ ] You have added `contentDescription` tags to new Views
- [ ] You have added sufficient documentation (descriptive comments).


app/src/main/java/piuk/blockchain/android/coincore/eth/EthAsset.kt
```diff
@@ -108,19 +110,26 @@ internal class EthAsset(
...
+                parts = parts[1].split("&");
+            }
+
+            val amountPart = parts.find {
+                it.startsWith(amountField, true)
+            }?.let {
+                CryptoValue.fromMinor(CryptoCurrency.ETHER, it.substring(amountField.length).toBigDecimal())
+            }
+            if (addressPart != null && isValidAddress(addressPart)) {
+                EthAddress(address = addressPart, label = label ?: address, amount = amountPart)
```
pauls-bc, line 130: We still need to do the isContractCheck() validation ATM, because it has implications elsewhere in the code wrt fee schedules; I have an open ticket to fix that, and then this can go, but just now it'll need to stay.

pauls-bc (commented):
lgtm. Just that validation check, but otherwise good. That amount field could probably move up the hierarchy into CryptoTarget, maybe. And its value will be ignored in the Tx flow as it stands, but that can be later, NP.

